PR #1, opened 2025-07-26
FEATURE: automatically downsize large images

everettbu:
Test 1

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js
+++ b/app/assets/javascripts/discourse/lib/utilities.js
@@ -179,7 +179,7 @@ Discourse.Utilities = {
 
     // check file size
     var fileSizeKB = file.size / 1024;
-    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
+    var maxSizeKB = 10 * 1024; // 10MB
     if (fileSizeKB > maxSizeKB) {
       bootbox.alert(I18n.t('post.errors.file_too_large', { max_size_kb: maxSizeKB }));
       return false;
@@ -243,7 +243,7 @@ Discourse.Utilities = {
 
         // entity too large, usually returned from the web server
         case 413:
-          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
+          var maxSizeKB = 10 * 1024; // 10 MB
           bootbox.alert(I18n.t('post.errors.file_too_large', { max_size_kb: maxSizeKB }));
           return;
 
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -52,12 +52,21 @@ class UploadsController < ApplicationController
     begin
       # API can provide a URL
       if file.nil? && url.present? && is_api?
-        tempfile = FileHelper.download(url, SiteSetting.max_image_size_kb.kilobytes, "discourse-upload-#{type}") rescue nil
+        tempfile = FileHelper.download(url, 10.megabytes, "discourse-upload-#{type}") rescue nil
         filename = File.basename(URI.parse(url).path)
       else
         tempfile = file.tempfile
         filename = file.original_filename
         content_type = file.content_type
+      end
+
+      # allow users to upload large images that will be automatically reduced to allowed size
+      if tempfile && tempfile.size > 0 && SiteSetting.max_image_size_kb > 0 && FileHelper.is_image?(filename)
+        attempt = 5
+        while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
+          OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
+          attempt -= 1
+        end
       end
 
       upload = Upload.create_for(current_user.id, tempfile, filename, tempfile.size, content_type: content_type, image_type: type)
--- a/app/models/optimized_image.rb
+++ b/app/models/optimized_image.rb
@@ -139,23 +139,22 @@ class OptimizedImage < ActiveRecord::Base
   end
 
   def self.resize(from, to, width, height, opts={})
-    optimize("resize", from, to, width, height, opts)
+    optimize("resize", from, to, "#{width}x#{height}", opts)
   end
 
   def self.downsize(from, to, max_width, max_height, opts={})
-    optimize("downsize", from, to, max_width, max_height, opts)
+    optimize("downsize", from, to, "#{max_width}x#{max_height}", opts)
   end
 
-  def self.optimize(operation, from, to, width, height, opts={})
-    dim = dimensions(width, height)
+  def self.downsize(from, to, dimensions, opts={})
+    optimize("downsize", from, to, dimensions, opts)
+  end
+
+  def self.optimize(operation, from, to, dimensions, opts={})
     method_name = "#{operation}_instructions"
     method_name += "_animated" if !!opts[:allow_animation] && from =~ /\.GIF$/i
-    instructions = self.send(method_name.to_sym, from, to, dim, opts)
+    instructions = self.send(method_name.to_sym, from, to, dimensions, opts)
     convert_with(instructions, to)
-  end
-
-  def self.dimensions(width, height)
-    "#{width}x#{height}"
   end
 
   def self.convert_with(instructions, to)


mfeuerstein (commented):
## PR Review — approved

Reviewed 3 files. 0 high-severity issues found. Verdict: approved.

### `app/assets/javascripts/discourse/lib/utilities.js` (low)

- Reviewed app/assets/javascripts/discourse/lib/utilities.js — looks good

### `app/models/optimized_image.rb` (low)

- Reviewed app/models/optimized_image.rb — looks good

### `app/controllers/uploads_controller.rb` (low)

- Reviewed app/controllers/uploads_controller.rb — looks good

app/models/optimized_image.rb
```diff
@@ -139,25 +139,24 @@ def self.downsize_instructions_animated(from, to, dimensions, opts={})
...
 
   def self.resize(from, to, width, height, opts={})
-    optimize("resize", from, to, width, height, opts)
+    optimize("resize", from, to, "#{width}x#{height}", opts)
   end
 
   def self.downsize(from, to, max_width, max_height, opts={})
-    optimize("downsize", from, to, max_width, max_height, opts)
+    optimize("downsize", from, to, "#{max_width}x#{max_height}", opts)
   end
```
zach-source, line 147: **[critical] Duplicate `downsize` definition — original 4-arg overload is silently replaced, breaking all existing callers**

Ruby uses the last definition when the same method name appears twice. The diff adds a new 3-arg `def self.downsize(from, to, dimensions, opts={})` but the prior `def self.downsize(from, to, max_width, max_height, opts={})` was already rewritten in the same hunk (line 143) to delegate to `optimize` with `"#{max_width}x#{max_height}"`. The net result is two `self.downsize` definitions: the one at ~143 (4-arg) and the new one at ~147 (3-arg). In Ruby the 3-arg version wins, so any caller still passing `downsize(from, to, width, height)` — including the now-modified `resize`-path callers anywhere else in the codebase — will silently pass a numeric `max_width` as `dimensions`, producing a malformed ImageMagick argument and a silent false return from `convert_with`.

**Suggestion:** Remove the duplicate 4-arg `downsize` definition entirely; keep only the 3-arg form and update all call sites to pre-format the dimension string before calling.

app/controllers/uploads_controller.rb
```diff
@@ -52,14 +52,23 @@ def create_upload(type, file, url)
     begin
       # API can provide a URL
       if file.nil? && url.present? && is_api?
-        tempfile = FileHelper.download(url, SiteSetting.max_image_size_kb.kilobytes, "discourse-upload-#{type}") rescue nil
+        tempfile = FileHelper.download(url, 10.megabytes, "discourse-upload-#{type}") rescue nil
```
zach-source, line 55: **[high] URL-sourced download limit hardcoded to 10 MB, ignoring `SiteSetting.max_image_size_kb`**

The original code used `SiteSetting.max_image_size_kb.kilobytes` as the download size cap for API URL uploads. The diff replaces it with the literal `10.megabytes`. This means admins who configure `max_image_size_kb` to a value smaller than 10 MB (e.g., 1 MB) will have that limit silently bypassed for API uploads — an API caller can now download and store images up to 10 MB regardless of site policy. It also means `FileHelper.download` fetches up to 10 MB before the downsize loop even runs, wasting bandwidth for non-image types that are not subject to the downsize path.

**Suggestion:** Keep using `SiteSetting.max_image_size_kb.kilobytes` as the download cap, or use `[SiteSetting.max_image_size_kb.kilobytes, 10.megabytes].max` if the intent is to allow downloading larger files for subsequent downsize.

```suggestion
        tempfile = FileHelper.download(url, [SiteSetting.max_image_size_kb.kilobytes, 10.megabytes].max, "discourse-upload-#{type}") rescue nil
```

app/assets/javascripts/discourse/lib/utilities.js
```diff
@@ -179,7 +179,7 @@ Discourse.Utilities = {
 
     // check file size
     var fileSizeKB = file.size / 1024;
-    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
+    var maxSizeKB = 10 * 1024; // 10MB
```
zach-source, line 182: **[high] Client-side size check hardcoded to 10 MB, bypassing per-type site settings**

The original code used `Discourse.SiteSettings['max_' + type + '_size_kb']` to enforce per-type limits (image vs. attachment). The diff replaces this with a hardcoded `10 * 1024` (10 MB) for all types. This means: (1) non-image file types (attachments) that have a site setting smaller than 10 MB are no longer client-side validated — users see no error until the server rejects them; (2) the `maxSizeKB` value passed to the `file_too_large` i18n message will always show "10240 KB" regardless of the actual configured limit, confusing users.

**Suggestion:** Restore the dynamic lookup: `var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];` — or, if the intent is to allow large image uploads for server-side downsize, special-case only image types.

```suggestion
    var maxSizeKB = type === 'image' ? 10 * 1024 : Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/assets/javascripts/discourse/lib/utilities.js
```diff
@@ -243,7 +243,7 @@ Discourse.Utilities = {
 
         // entity too large, usually returned from the web server
         case 413:
-          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
+          var maxSizeKB = 10 * 1024; // 10 MB
```
zach-source, line 246: **[medium] 413 error handler shows hardcoded 10 MB instead of actual server limit**

The 413 (Request Entity Too Large) handler is supposed to tell the user the configured maximum. By hardcoding `10 * 1024` here, any site with `max_image_size_kb` set differently will show the wrong limit in the error dialog. This is misleading: the server rejected the upload based on its configured limit, but the client reports a different number.

**Suggestion:** Restore `var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;`

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

zach-source (changes requested):
Found 4 blocking issues.

- **critical** `app/models/optimized_image.rb:147` — Duplicate `downsize` definition — original 4-arg overload is silently replaced, breaking all existing callers
- **high** `app/controllers/uploads_controller.rb:55` — URL-sourced download limit hardcoded to 10 MB, ignoring `SiteSetting.max_image_size_kb`
- **high** `app/assets/javascripts/discourse/lib/utilities.js:182` — Client-side size check hardcoded to 10 MB, bypassing per-type site settings
- **medium** `app/assets/javascripts/discourse/lib/utilities.js:246` — 413 error handler shows hardcoded 10 MB instead of actual server limit

app/models/optimized_image.rb
```diff
@@ -139,25 +139,24 @@ def self.downsize_instructions_animated(from, to, dimensions, opts={})
...
```
zach-source, line 148: Ruby method overloading: second `downsize` replaces the first. All four-arg callers will fail silently.

Remove the old definition entirely; keep only the new 3-argument version.

app/controllers/uploads_controller.rb
```diff
@@ -52,14 +52,23 @@ def create_upload(type, file, url)
...
       else
         tempfile = file.tempfile
         filename = file.original_filename
         content_type = file.content_type
       end
 
+      # allow users to upload large images that will be automatically reduced to allowed size
+      if tempfile && tempfile.size > 0 && SiteSetting.max_image_size_kb > 0 && FileHelper.is_image?(filename)
+        attempt = 5
+        while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
```
zach-source, line 66: Return value ignored: if all `downsize` calls fail, loop exhausts attempts and uploads oversized file anyway.

```suggestion:-0+1
          break unless OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
          attempt -= 1
```

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 182: Hardcoded 10MB bypasses admin-configured limits for non-image uploads.

```suggestion
    var maxSizeKB = (type === 'image') ? (10 * 1024) : Discourse.SiteSettings['max_' + type + '_size_kb'];
```

```suggestion
    var maxSizeKB = (type === 'image') ? 10 * 1024 : Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 246: Error message shows hardcoded 10MB instead of actual admin-configured limit.

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

zach-source (changes requested):
Method overloading, missing error handling, hardcoded upload limits.

app/models/optimized_image.rb
```diff
@@ -139,25 +139,24 @@ def self.downsize_instructions_animated(from, to, dimensions, opts={})
...
 
-  def self.optimize(operation, from, to, width, height, opts={})
-    dim = dimensions(width, height)
+  def self.downsize(from, to, dimensions, opts={})
```
zach-source, line 149: Method `downsize` defined twice — 3-arg version shadows the 4-arg version at line 145. Remove the redundant 4-arg method (lines 145–147) since both convert dimensions the same way.

app/controllers/uploads_controller.rb
zach-source, line 66: Image re-encoding can increase file size (format overhead). Loop runs 5 attempts regardless of shrinkage, then uploads oversized file. Track previous size and break if unchanged:

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 182: Hardcoded 10 MB check ignores `max_image_size_kb` site setting. Sites with lower limits bypass validation entirely. Read the site setting instead:

```suggestion
    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 246: Error message hardcoded to 10240 KB, ignoring actual `max_image_size_kb` setting. Users see wrong limit if configured differently. Restore the site-setting read:

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

app/controllers/uploads_controller.rb
zach-source, line 55: Download cap hardcoded to 10 MB ignores `max_image_size_kb` setting. Wastes bandwidth on small-limit sites, truncates valid uploads on large-limit sites. Use proportional cap:

```suggestion
        tempfile = FileHelper.download(url, [SiteSetting.max_image_size_kb.kilobytes * 5, 10.megabytes].max, "discourse-upload-#{type}") rescue nil
```

zach-source (changes requested):
Method shadowing causes parameter mismatch; resize loop proceeds even when images don't shrink; size limits hardcoded throughout, ignoring site settings — see inline.

app/models/optimized_image.rb
zach-source, line 148: Duplicate `downsize` method—line 145 already defines it. Ruby keeps only the last definition (line 148), making the 4-arg version unreachable and breaking existing callers. Remove lines 145–147.

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 182: Hard-coded 10 MB override ignores per-file-type site settings. Admins' `max_image_size_kb` and `max_attachment_size_kb` now ignored. Restore:

```suggestion
    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
```

```suggestion
    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'] || 10 * 1024;
```

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 246: Hard-coded error message (10240 KB) hides actual server limit, confusing users if different. Use:

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb || 10 * 1024;
```

app/models/optimized_image.rb
```diff
@@ -139,25 +139,24 @@ def self.downsize_instructions_animated(from, to, dimensions, opts={})
   end
 
   def self.resize(from, to, width, height, opts={})
-    optimize("resize", from, to, width, height, opts)
+    optimize("resize", from, to, "#{width}x#{height}", opts)
   end
 
   def self.downsize(from, to, max_width, max_height, opts={})
```
zach-source, line 145: Dead-code method—unreachable since line 148 redefines `downsize` with a different arity (3 args vs 4 args). Delete:

```suggestion:-0+2

```

zach-source (changes requested):
Method redefinition creates runtime errors; hard-coded limits override site settings; confusing error message. See inline.

app/models/optimized_image.rb
zach-source, line 147: Two `downsize` methods with same name — Ruby uses last definition, breaking old 4-arg callers. Delete the old definition and update all callers to pre-format dimensions as "WIDTHxHEIGHT" string (or percentage like "80%").

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 182: Hardcoded 10 MB limit replaces per-type server setting. Error message shows wrong limit, breaks client-side validation for non-image types. Use `Discourse.SiteSettings['max_' + type + '_size_kb']`:

```suggestion
    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/assets/javascripts/discourse/lib/utilities.js
zach-source, line 246: Hardcoded 10 MB limit replaces `max_image_size_kb` setting. Error message shows wrong limit if admin configured lower value. Use `Discourse.SiteSettings.max_image_size_kb`:

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

zach-source (changes requested):
Method collision and hardcoded limits flagged inline.

app/models/optimized_image.rb
ron-x5labs, line 149: **[blocking] Duplicate `downsize` definition shadows the 5-arg version, breaking existing callers**

Ruby has no method overloading by arity — the second `def self.downsize(from, to, dimensions, opts={})` (4 params) completely replaces the 5-param `def self.downsize(from, to, max_width, max_height, opts={})` defined just above it. The last definition wins.

The existing caller `app/jobs/regular/resize_emoji.rb:14` invokes `OptimizedImage.downsize(path, path, 100, 100, opts)` with 5 positional arguments, which now raises `ArgumentError: wrong number of arguments (given 5, expected 3..4)` and crashes every `Jobs::ResizeEmoji` run.

Fix: keep a single method that accepts either form, e.g.

```ruby
def self.downsize(from, to, *args)
  opts = args.last.is_a?(Hash) ? args.pop : {}
  dims = args.size == 2 ? "#{args[0]}x#{args[1]}" : args[0]
  optimize("downsize", from, to, dims, opts)
end
```

…or update `resize_emoji.rb` (and any other 5-arg callers) to pass a pre-formatted dimensions string and delete the dead 5-param definition.

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 182: **[blocking] Hardcoded 10 MB replaces per-type site settings and breaks an existing test**

The original `Discourse.SiteSettings['max_' + type + '_size_kb']` enforced admin-configured `max_image_size_kb` / `max_attachment_size_kb` per type. Replacing it with `10 * 1024` (10 MB) for all types means:
1. Admin limits are bypassed client-side — attachments configured below 10 MB are no longer validated before upload, so users hit a server rejection with a contradictory limit shown.
2. The error dialog always reports "10240 KB" regardless of the actual configured limit.

This also **breaks the existing test** `test/javascripts/lib/utilities-test.js.es6:52` ("prevents files that are too big from being uploaded"), which sets `max_image_size_kb = 5` and expects a 10 KB file to be rejected with `max_size_kb: 5`. With the hardcoded `10 * 1024` KB, that 10 KB file passes validation and both assertions fail.

Restore the dynamic lookup:

```suggestion
    var maxSizeKB = Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/controllers/uploads_controller.rb
ron-x5labs, line 66: **[blocking] Downsize loop ignores failures and has no progress detection**

`OptimizedImage.downsize` → `convert_with` returns `false` on ImageMagick failure (`$?.exitstatus != 0`) or when `ImageOptim` raises. This loop never checks the return value, so:
- If `convert`/`gifsicle` is missing or fails, the loop spins 5 times doing nothing, then the unchanged oversized (or corrupt) file is passed to `Upload.create_for` on line 72 and rejected with a generic "file too large" error — no log explains that auto-resize failed.
- Re-encoding an already-compressed image can *increase* size (format/metadata overhead); the loop has no way to detect a stall and will exhaust all 5 attempts, then upload an oversized file that the validator rejects.

Track the previous size, check the return value, and stop early:

```ruby
attempt = 5
prev_size = tempfile.size
while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
  ok = OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
  break if !ok || tempfile.size >= prev_size  # downsize failed or made no progress
  prev_size = tempfile.size
  attempt -= 1
end
Rails.logger.warn("Failed to downsize upload below max_image_size_kb after #{5 - attempt} attempts") if tempfile.size > SiteSetting.max_image_size_kb.kilobytes
```

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 246: **[non-blocking] 413 handler shows a hardcoded 10 MB limit**

A 413 (Request Entity Too Large) comes from the web server's `client_max_body_size`, which the `max_image_size_kb` setting description tells admins to keep in sync. Hardcoding `10 * 1024` here means the dialog reports a 10 MB limit even when the server is configured differently, misleading the user. Restore the site setting so the message matches reality:

```suggestion
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

app/controllers/uploads_controller.rb
ron-x5labs, line 55: **[non-blocking] Download cap hardcoded to 10 MB bypasses `max_image_size_kb`**

The original `SiteSetting.max_image_size_kb.kilobytes` bounded API URL downloads to the admin-configured limit. Replacing it with `10.megabytes` means:
- Sites with a low `max_image_size_kb` still fetch up to 10 MB per API upload before the downsize loop runs — wasted bandwidth and temp storage.
- For non-image attachments (which skip the downsize loop entirely), this is the only pre-`create_for` bound, so a 10 MB fetch bypasses the admin's intended attachment policy until the validator rejects it.

If the intent is to allow headroom for images that will be downsized, derive the cap from the setting rather than a magic number, e.g. `[SiteSetting.max_image_size_kb.kilobytes * 5, 10.megabytes].max`.

app/controllers/uploads_controller.rb
```diff
@@ -52,14 +52,23 @@ def create_upload(type, file, url)
...
+          OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
```
ron-x5labs, line 67: **[non-blocking] In-place downsize can corrupt the sole upload copy on a mid-write crash**

`OptimizedImage.downsize(tempfile.path, tempfile.path, ...)` uses the same path as both source and destination; `convert_with` invokes an external `convert`/`gifsicle` process that writes directly to that path. If the process is killed or the disk fills mid-write, the downloaded tempfile — the only copy of the user's upload — is left truncated or partially written, and the `return false if $?.exitstatus != 0` check can't undo the damage. Downsize to a temporary output file and atomically rename on success.

ron-x5labs, line 67: **[non-blocking] `ImageOptim.new` is re-instantiated on every `convert_with` call**

`convert_with` (`optimized_image.rb:164`) constructs `ImageOptim.new` each invocation, which performs on-disk binary discovery (jpegoptim, optipng, …). This 5-iteration loop triggers up to 5 redundant discovery scans per oversized upload. Memoize the `ImageOptim` instance (e.g. a class-level `@@image_optim ||= ImageOptim.new`) or accept it as a parameter.

ron-x5labs (changes requested):
## Code Review: FEATURE: automatically downsize large images

### Problem
The PR attempts to let users upload large images by automatically downsizing them server-side before storing, so admins can keep a small `max_image_size_kb` without rejecting oversized-but-valid images.

### Solution Reviewed
Three changes: (1) `OptimizedImage` refactored so `downsize`/`resize` pass a pre-formatted dimensions string (`"WxH"` or `"80%"`) into `optimize` instead of separate width/height args; (2) `UploadsController#create_upload` adds a loop that repeatedly calls `OptimizedImage.downsize(tempfile, tempfile, "80%", ...)` up to 5 times when an uploaded image exceeds `max_image_size_kb`; (3) client-side size checks in `utilities.js` and the API download cap switched from site settings to a hardcoded 10 MB.

### Summary
The core idea is reasonable, but the implementation has a crashing regression (duplicate method definition shadows the old arity and breaks existing callers), silently ignores downsize failures, and hardcodes 10 MB in three places that bypass admin-configured per-type limits and break an existing test. Needs changes before merge.

### Files Reviewed
- `app/models/optimized_image.rb` — deeply reviewed
- `app/controllers/uploads_controller.rb` — deeply reviewed
- `app/assets/javascripts/discourse/lib/utilities.js` — deeply reviewed
- `app/jobs/regular/resize_emoji.rb` — deeply reviewed (existing caller, not in diff)
- `test/javascripts/lib/utilities-test.js.es6` — reviewed for test impact (not in diff)
- `CONTRIBUTING.md` — lightly reviewed (context freshness)

### Verification
- `ruby -c` — skipped: no Ruby interpreter available in the review environment; static analysis used instead.
- `OptimizedImage.downsize` shadowing → `ArgumentError` for 5-arg callers: confirmed by reading `resize_emoji.rb:14` (`downsize(path, path, 100, 100, opts)`) against the new 4-param definition.
- Existing JS test break: confirmed by reading `utilities-test.js.es6:52-61` — sets `max_image_size_kb=5`, expects a 10 KB file rejected with `max_size_kb:5`; the hardcoded `10*1024` KB makes that file pass, failing both assertions.
- `convert_with` return-value/`ImageOptim.new` behavior: confirmed at `optimized_image.rb:160-169`.

### Issues Found
See inline comments. Additional non-inline notes:

- **Missing test coverage (non-blocking):** The new auto-downsize loop in `uploads_controller.rb` and the new 4-param `OptimizedImage.downsize` have no spec coverage in `spec/controllers/uploads_controller_spec.rb` or `spec/models/optimized_image_spec.rb`. Add a test for an oversized image upload that exercises the loop, and one for the `ArgumentError` regression path.
- **Stale context file (suggestion):** `CONTRIBUTING.md` references the unauthenticated `git://` clone protocol ("git clone git://github.com/discourse/discourse.git", line 33) — GitHub disabled this in 2021 — and links to `try.discourse.org` (line 13) which may no longer be maintained. Update to `https://` and verify the sandbox link.

### Verdict
Recommend changes before merge — the duplicate `downsize` definition crashes the existing emoji-resize job, the downsize loop ignores failures, and the hardcoded 10 MB limits bypass admin settings and break an existing client-side test.

app/models/optimized_image.rb
ron-x5labs, line 149: **[blocking] Duplicate `downsize` definition shadows the 5-arg overload, breaking the existing `ResizeEmoji` job**

Ruby has no overloading by arity — the last definition wins. The new `def self.downsize(from, to, dimensions, opts={})` (4 params) completely replaces the `def self.downsize(from, to, max_width, max_height, opts={})` at line 145. The existing caller `app/jobs/regular/resize_emoji.rb:14` calls `OptimizedImage.downsize(path, path, 100, 100, opts)` with 5 positional args, so every emoji resize now raises `ArgumentError: wrong number of arguments (given 5, expected 3..4)`. No spec references `downsize`, so this is not caught by the suite.

Delete the dead 4-arg def and migrate the caller to the dimensions-string form:
```ruby
# app/models/optimized_image.rb — keep only:
def self.downsize(from, to, dimensions, opts={})
  optimize("downsize", from, to, dimensions, opts)
end

# app/jobs/regular/resize_emoji.rb:14
OptimizedImage.downsize(path, path, "100x100", opts)
```

app/controllers/uploads_controller.rb
ron-x5labs, line 67: **[blocking] Downsize loop ignores the return value and mutates the sole copy in place**

`OptimizedImage.downsize` → `convert_with` returns `false` on a non-zero `convert`/ImageOptim exit or any rescued exception. This loop never checks that return, so if `convert`/`gifsicle` is missing or the input is corrupt, the loop spins 5 times doing nothing and falls through to `Upload.create_for` with a possibly-truncated/0-byte tempfile. Because `from == to` (`tempfile.path`), a failed or interrupted `convert` corrupts the only copy of the uploaded/downloaded file with no fallback; the `UploadValidator` only enforces an upper size bound, so an empty/corrupt file may be stored.

Check the return and bail out on failure:
```ruby
attempt = 5
while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
  break unless OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
  attempt -= 1
end
```
Consider also re-checking `tempfile.size` after the loop and rejecting (returning an errors payload) if it is still over the limit.

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 182: **[blocking] Hardcoded 10 MB replaces per-type admin settings, bypassing the limit for non-image attachments**

The original `Discourse.SiteSettings['max_' + type + '_size_kb']` enforced admin-configured `max_image_size_kb` / `max_attachment_size_kb` per type. Hardcoding `10 * 1024` for all types means attachments configured below 10 MB get no client-side gate, and the server-side downsize loop only runs for images (`FileHelper.is_image?`), so oversized non-image files are submitted with no client check and no reduction. This is a config-contract break — admins can no longer tune upload limits per type.

Keep the admin setting for non-image types and only widen the cap for images (which the server will downsize):
```js
var maxSizeKB = (type === 'image') ? 10 * 1024 : Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 246: **[non-blocking] 413 handler shows a hardcoded 10 MB instead of the actual server limit**

A 413 comes from the web server's `client_max_body_size`, which the `max_image_size_kb` setting description tells admins to keep in sync. Hardcoding `10 * 1024` here reports a 10 MB limit even when the server is configured differently, misleading the user about the real constraint. Restore the site setting so the message matches:
```js
var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

app/controllers/uploads_controller.rb
ron-x5labs, line 55: **[non-blocking] Download cap raised to a fixed 10 MB widens pre-existing SSRF and wastes bandwidth on low-limit sites**

`FileHelper.download` only validates `url =~ /^https?\/\/` (no private/loopback/link-local host block), so API URL uploads can already reach internal endpoints (e.g. `169.254.169.254` cloud metadata). Raising the cap from `SiteSetting.max_image_size_kb.kilobytes` to a fixed `10.megabytes` increases how much data can be pulled from internal hosts per request, and forces low-limit sites to fetch up to 10 MB before the downsize loop runs. Use a cap proportional to the admin setting:
```ruby
tempfile = FileHelper.download(url, [SiteSetting.max_image_size_kb.kilobytes * 5, 10.megabytes].max, "discourse-upload-#{type}") rescue nil
```
(Separately, adding an SSRF guard that rejects resolved IPs in private/loopback ranges in `FileHelper.download` would close the pre-existing hole.)

app/controllers/uploads_controller.rb
ron-x5labs, line 66: **[non-blocking] No progress detection — the loop can burn 5 full `convert` + `ImageOptim` passes on a file that never shrinks**

If `downsize "80%"` doesn't reduce byte size (tiny-dimension image, metadata/EXIF-heavy file, PNG that re-encodes larger, or already at minimum), the loop re-runs the expensive `convert` + `ImageOptim.new.optimize_image!` pipeline 5 times producing the same file, wasting CPU/disk — and on the now-permitted 10 MB inputs this is a meaningful DoS amplification. Track the previous size and exit early when it stops changing:
```ruby
attempt = 5
prev_size = nil
while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
  break if tempfile.size == prev_size
  prev_size = tempfile.size
  break unless OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
  attempt -= 1
end
```

ron-x5labs (changes requested):
## Code Review: FEATURE: automatically downsize large images

### Problem
Let users upload large images by automatically downsizing them server-side before storing, so admins can keep a small `max_image_size_kb` without rejecting oversized-but-valid images.

### Solution Reviewed
Adds a server-side 5-iteration downsize loop in `UploadsController#create_upload` that calls `OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", ...)` in place until the tempfile is under `max_image_size_kb`. Refactors `OptimizedImage#optimize` to take a single dimensions string instead of `(width, height)`, and widens the client-side and API-download size caps to a hardcoded 10 MB.

### Summary
The core auto-downsize idea is sound, but the refactor introduces a Ruby method-redefinition bug that breaks the existing `ResizeEmoji` job (immediate `ArgumentError`), the downsize loop ignores failures and can corrupt the sole copy of the file, and hardcoding 10 MB in three places silently overrides admin-configured per-type upload limits. Needs changes before merge.

### Files Reviewed
- `app/models/optimized_image.rb` — deeply reviewed (signature refactor + duplicate `downsize` def)
- `app/controllers/uploads_controller.rb` — deeply reviewed (new downsize loop + download cap)
- `app/assets/javascripts/discourse/lib/utilities.js` — deeply reviewed (client-side size gates)
- `app/jobs/regular/resize_emoji.rb` — deeply reviewed (existing `downsize` caller, not in diff but impacted)
- `config/site_settings.yml` / `lib/file_helper.rb` — lightly reviewed (context for the limits and download path)

### Verification
- `ruby` toolchain not installed in this environment — typecheck/test run skipped.
- Cross-file impact verified via `grep`: `app/jobs/regular/resize_emoji.rb:14` calls `OptimizedImage.downsize(path, path, 100, 100, opts)` (5 args) against the new 4-param winning definition.
- Spec coverage checked: zero specs reference `downsize` (`spec/controllers/uploads_controller_spec.rb`, `spec/models/optimized_image_spec.rb`), so the signature break is not caught by the suite.

### Verdict
Recommend changes before merge — the duplicate `downsize` definition is a runtime-breaking regression for the emoji job, the loop can corrupt the upload in place, and the hardcoded limits override admin config.

app/models/optimized_image.rb
ron-x5labs, line 149: **[blocking] Duplicate `def self.downsize` shadows the 5-arg overload, crashing `Jobs::ResizeEmoji`**

Ruby has no method overloading by arity — the last `def` wins. The new `def self.downsize(from, to, dimensions, opts={})` at line 149 completely replaces the `def self.downsize(from, to, max_width, max_height, opts={})` at line 145, so the 5-arg version is unreachable. The existing caller `app/jobs/regular/resize_emoji.rb:14` invokes `OptimizedImage.downsize(path, path, 100, 100, opts)` with 5 positional args, which now raises `ArgumentError: wrong number of arguments (given 5, expected 3..4)` on every emoji resize (no spec references `downsize`, so CI stays green while production breaks).

Delete the dead 5-arg def and migrate the emoji caller to the dimensions-string form:

```ruby
# app/jobs/regular/resize_emoji.rb
OptimizedImage.downsize(path, path, "100x100", opts)
```
```ruby
# app/models/optimized_image.rb — keep only one definition:
def self.downsize(from, to, dimensions, opts={})
  optimize("downsize", from, to, dimensions, opts)
end
```

app/controllers/uploads_controller.rb
ron-x5labs, line 67: **[blocking] Downsize loop ignores the `downsize` return value and has no stall detection**

`OptimizedImage.downsize` -> `convert_with` (optimized_image.rb:160-169) returns `false` on a non-zero `convert`/`gifsicle` exit or any rescued exception. This loop never checks that return, so if `convert` is missing, the input is corrupt, or re-encoding grows the file (PNG high-entropy, EXIF-heavy, already-minimal), the loop spins 5x producing the same output, then passes the unchanged/oversized tempfile to `Upload.create_for`. On the now-permitted 10 MB inputs that is 5 wasted `convert` + `ImageOptim.new` passes per oversized upload — a real CPU amplification. Because `from == to` (`tempfile.path`), a failed or interrupted `convert` can also truncate the only copy of the upload with no rollback.

Check the return and bail out on failure; track the previous size to break on a stall:

```ruby
attempt = 5
prev_size = nil
while attempt > 0 && tempfile.size > SiteSetting.max_image_size_kb.kilobytes
  break if tempfile.size == prev_size          # no progress -> stop spinning
  prev_size = tempfile.size
  ok = OptimizedImage.downsize(tempfile.path, tempfile.path, "80%", allow_animation: SiteSetting.allow_animated_thumbnails)
  break unless ok                               # convert failed -> don't keep retrying
  attempt -= 1
end
```

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 182: **[blocking] Hardcoded 10 MB replaces per-type site settings for all upload types and breaks an existing test**

`type` is `image` or `attachment`, and the original `Discourse.SiteSettings['max_' + type + '_size_kb']` enforced the admin-configured limit per type. Hardcoding `10 * 1024` for all types means: (1) non-image attachments configured below 10 MB get no client-side gate — the server still rejects them via `Validators::UploadValidator#maximum_file_size`, but only after a wasted full upload; (2) the `file_too_large` dialog always reports "10240 KB" regardless of the configured limit, misleading users. It also breaks the existing test `test/javascripts/lib/utilities-test.js.es6:52` ("prevents files that are too big"), which sets `max_image_size_kb = 5` and asserts the alert shows `max_size_kb: 5` — with the hardcoded 10240 the 10 KB test file passes validation, failing both assertions.

Widen the cap only for images (which the server downsizes) and keep the per-type setting otherwise:

```js
var maxSizeKB = (type === 'image') ? (10 * 1024) : Discourse.SiteSettings['max_' + type + '_size_kb'];
```

app/controllers/uploads_controller.rb
ron-x5labs, line 55: **[non-blocking] Download cap hardcoded to 10 MB bypasses `max_image_size_kb` and amplifies pre-existing SSRF**

The original `SiteSetting.max_image_size_kb.kilobytes` bounded API URL downloads to the admin-configured limit. `FileHelper.download` (file_helper.rb:10) validates only `url =~ /^https?:\/\//` with no private/loopback/link-local host block, so API URL uploads can already reach internal hosts (e.g. cloud metadata `169.254.169.254`) — this is pre-existing SSRF, and raising the cap to a fixed 10 MB increases how much data an API caller can pull from internal hosts per request. It also forces low-limit sites to fetch up to 10 MB before the downsize loop runs, and for non-image API uploads (which skip the loop) the 10 MB is fetched then likely rejected — wasted bandwidth. Derive the cap from the setting instead:

```ruby
tempfile = FileHelper.download(url, [SiteSetting.max_image_size_kb.kilobytes * 5, 10.megabytes].max, "discourse-upload-#{type}") rescue nil
```

app/assets/javascripts/discourse/lib/utilities.js
ron-x5labs, line 246: **[non-blocking] 413 handler reports a hardcoded 10 MB instead of the configured limit**

A 413 (Request Entity Too Large) comes from the web server's `client_max_body_size`, which the `max_image_size_kb` setting description tells admins to keep in sync. Hardcoding `10 * 1024` here reports a 10 MB limit even when the server is configured differently, misleading the user about the real constraint. Restore the site setting so the message matches reality:

```js
          var maxSizeKB = Discourse.SiteSettings.max_image_size_kb;
```

PR #4, opened 2025-07-26
Enhance embed URL handling and validation system

everettbu:
Test 4

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -206,6 +206,10 @@ gem 'unicorn', require: false
 gem 'puma', require: false
 gem 'rbtrace', require: false
 
+# required for feed importing and embedding
+gem 'ruby-readability', require: false
+gem 'simple-rss', require: false
+
 # perftools only works on 1.9 atm
 group :profile do
   # travis refuses to install this, instead of fuffing, just avoid it for now
--- a/app/assets/javascripts/embed.js
+++ b/app/assets/javascripts/embed.js
@@ -1,0 +1,27 @@
+/* global discourseUrl */
+/* global discourseEmbedUrl */
+(function() {
+
+  var comments = document.getElementById('discourse-comments'),
+      iframe = document.createElement('iframe');
+  iframe.src = discourseUrl + "embed/best?embed_url=" + encodeURIComponent(discourseEmbedUrl);
+  iframe.id = 'discourse-embed-frame';
+  iframe.width = "100%";
+  iframe.frameBorder = "0";
+  iframe.scrolling = "no";
+  comments.appendChild(iframe);
+
+
+  function postMessageReceived(e) {
+    if (!e) { return; }
+    if (discourseUrl.indexOf(e.origin) === -1) { return; }
+
+    if (e.data) {
+      if (e.data.type === 'discourse-resize' && e.data.height) {
+        iframe.height = e.data.height + "px";
+      }
+    }
+  }
+  window.addEventListener('message', postMessageReceived, false);
+
+})();
--- a/app/controllers/embed_controller.rb
+++ b/app/controllers/embed_controller.rb
@@ -1,0 +1,34 @@
+class EmbedController < ApplicationController
+  skip_before_filter :check_xhr
+  skip_before_filter :preload_json
+  before_filter :ensure_embeddable
+
+  layout 'embed'
+
+  def best
+    embed_url = params.require(:embed_url)
+    topic_id = TopicEmbed.topic_id_for_embed(embed_url)
+
+    if topic_id
+      @topic_view = TopicView.new(topic_id, current_user, {best: 5})
+    else
+      Jobs.enqueue(:retrieve_topic, user_id: current_user.try(:id), embed_url: embed_url)
+      render 'loading'
+    end
+
+    discourse_expires_in 1.minute
+  end
+
+  private
+
+    def ensure_embeddable
+      raise Discourse::InvalidAccess.new('embeddable host not set') if SiteSetting.embeddable_host.blank?
+      raise Discourse::InvalidAccess.new('invalid referer host') if URI(request.referer || '').host != SiteSetting.embeddable_host
+
+      response.headers['X-Frame-Options'] = "ALLOWALL"
+    rescue URI::InvalidURIError
+      raise Discourse::InvalidAccess.new('invalid referer host')
+    end
+
+
+end
--- a/app/jobs/scheduled/poll_feed.rb
+++ b/app/jobs/scheduled/poll_feed.rb
@@ -1,0 +1,41 @@
+#
+# Creates and Updates Topics based on an RSS or ATOM feed.
+#
+require 'digest/sha1'
+require_dependency 'post_creator'
+require_dependency 'post_revisor'
+require 'open-uri'
+
+module Jobs
+  class PollFeed < Jobs::Scheduled
+    recurrence { hourly }
+    sidekiq_options retry: false
+
+    def execute(args)
+      poll_feed if SiteSetting.feed_polling_enabled? &&
+                   SiteSetting.feed_polling_url.present? &&
+                   SiteSetting.embed_by_username.present?
+    end
+
+    def feed_key
+      @feed_key ||= "feed-modified:#{Digest::SHA1.hexdigest(SiteSetting.feed_polling_url)}"
+    end
+
+    def poll_feed
+      user = User.where(username_lower: SiteSetting.embed_by_username.downcase).first
+      return if user.blank?
+
+      require 'simple-rss'
+      rss = SimpleRSS.parse open(SiteSetting.feed_polling_url)
+
+      rss.items.each do |i|
+        url = i.link
+        url = i.id if url.blank? || url !~ /^https?\:\/\//
+
+        content = CGI.unescapeHTML(i.content.scrub)
+        TopicEmbed.import(user, url, i.title, content)
+      end
+    end
+
+  end
+end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -60,6 +60,10 @@ class Post < ActiveRecord::Base
     @types ||= Enum.new(:regular, :moderator_action)
   end
 
+  def self.cook_methods
+    @cook_methods ||= Enum.new(:regular, :raw_html)
+  end
+
   def self.find_by_detail(key, value)
     includes(:post_details).where(post_details: { key: key, value: value }).first
   end
@@ -124,6 +128,11 @@ class Post < ActiveRecord::Base
   end
 
   def cook(*args)
+    # For some posts, for example those imported via RSS, we support raw HTML. In that
+    # case we can skip the rendering pipeline.
+    return raw if cook_method == Post.cook_methods[:raw_html]
+
+    # Default is to cook posts
     Plugin::Filter.apply(:after_post_cook, self, post_analyzer.cook(*args))
   end
 
--- a/app/models/topic_embed.rb
+++ b/app/models/topic_embed.rb
@@ -1,0 +1,82 @@
+require_dependency 'nokogiri'
+
+class TopicEmbed < ActiveRecord::Base
+  belongs_to :topic
+  belongs_to :post
+  validates_presence_of :embed_url
+  validates_presence_of :content_sha1
+
+  # Import an article from a source (RSS/Atom/Other)
+  def self.import(user, url, title, contents)
+    return unless url =~ /^https?\:\/\//
+
+    contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: "<a href='#{url}'>#{url}</a>")}</small>\n"
+
+    embed = TopicEmbed.where(embed_url: url).first
+    content_sha1 = Digest::SHA1.hexdigest(contents)
+    post = nil
+
+    # If there is no embed, create a topic, post and the embed.
+    if embed.blank?
+      Topic.transaction do
+        creator = PostCreator.new(user, title: title, raw: absolutize_urls(url, contents), skip_validations: true, cook_method: Post.cook_methods[:raw_html])
+        post = creator.create
+        if post.present?
+          TopicEmbed.create!(topic_id: post.topic_id,
+                             embed_url: url,
+                             content_sha1: content_sha1,
+                             post_id: post.id)
+        end
+      end
+    else
+      post = embed.post
+      # Update the topic if it changed
+      if content_sha1 != embed.content_sha1
+        revisor = PostRevisor.new(post)
+        revisor.revise!(user, absolutize_urls(url, contents), skip_validations: true, bypass_rate_limiter: true)
+        embed.update_column(:content_sha1, content_sha1)
+      end
+    end
+
+    post
+  end
+
+  def self.import_remote(user, url, opts=nil)
+    require 'ruby-readability'
+
+    opts = opts || {}
+    doc = Readability::Document.new(open(url).read,
+                                        tags: %w[div p code pre h1 h2 h3 b em i strong a img],
+                                        attributes: %w[href src])
+
+    TopicEmbed.import(user, url, opts[:title] || doc.title, doc.content)
+  end
+
+  # Convert any relative URLs to absolute. RSS is annoying for this.
+  def self.absolutize_urls(url, contents)
+    uri = URI(url)
+    prefix = "#{uri.scheme}://#{uri.host}"
+    prefix << ":#{uri.port}" if uri.port != 80 && uri.port != 443
+
+    fragment = Nokogiri::HTML.fragment(contents)
+    fragment.css('a').each do |a|
+      href = a['href']
+      if href.present? && href.start_with?('/')
+        a['href'] = "#{prefix}/#{href.sub(/^\/+/, '')}"
+      end
+    end
+    fragment.css('img').each do |a|
+      src = a['src']
+      if src.present? && src.start_with?('/')
+        a['src'] = "#{prefix}/#{src.sub(/^\/+/, '')}"
+      end
+    end
+
+    fragment.to_html
+  end
+
+  def self.topic_id_for_embed(embed_url)
+    TopicEmbed.where(embed_url: embed_url).pluck(:topic_id).first
+  end
+
+end
--- a/app/views/embed/best.html.erb
+++ b/app/views/embed/best.html.erb
@@ -1,0 +1,30 @@
+<header>
+  <%- if @topic_view.posts.present? %>
+    <%= link_to(I18n.t('embed.title'), @topic_view.topic.url, class: 'button', target: '_blank') %>
+  <%- else %>
+    <%= link_to(I18n.t('embed.start_discussion'), @topic_view.topic.url, class: 'button', target: '_blank') %>
+  <%- end if %>
+
+  <%= link_to(image_tag(SiteSetting.logo_url, class: 'logo'), Discourse.base_url) %>
+</header>
+
+<%- if @topic_view.posts.present? %>
+  <%- @topic_view.posts.each do |post| %>
+    <article class='post'>
+      <%= link_to post.created_at.strftime("%e %b %Y"), post.url, class: 'post-date', target: "_blank" %>
+      <div class='author'>
+        <img src='<%= post.user.small_avatar_url %>'>
+        <h3><%= post.user.username %></h3>
+      </div>
+      <div class='cooked'><%= raw post.cooked %></div>
+      <div style='clear: both'></div>
+    </article>
+  <%- end %>
+
+  <footer>
+    <%= link_to(I18n.t('embed.continue'), @topic_view.topic.url, class: 'button', target: '_blank') %>
+    <%= link_to(image_tag(SiteSetting.logo_url, class: 'logo'), Discourse.base_url) %>
+  </footer>
+
+<% end %>
+
--- a/app/views/embed/loading.html.erb
+++ b/app/views/embed/loading.html.erb
@@ -1,0 +1,12 @@
+<header>
+  <%= t 'embed.loading' %>
+  <%= link_to(image_tag(SiteSetting.logo_url, class: 'logo'), Discourse.base_url) %>
+</header>
+
+<script>
+  (function() {
+    setTimeout(function() {
+      document.location.reload();
+    }, 30000);
+  })();
+</script>
--- a/app/views/layouts/embed.html.erb
+++ b/app/views/layouts/embed.html.erb
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <%= stylesheet_link_tag 'embed' %>
+
+    <script>
+      (function() {
+        window.onload = function() {
+          if (parent) {
+            // Send a post message with our loaded height
+            parent.postMessage({type: 'discourse-resize', height: document['body'].offsetHeight}, '<%= request.referer %>');
+          }
+        }
+      })();
+    </script>
+  </head>
+  <body>
+    <%= yield %>
+  </body>
+</html>
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -29,6 +29,14 @@ en:
 
   too_many_replies: "Sorry you can't reply any more times in that topic."
 
+  embed:
+    title: "Discussion Highlights"
+    start_discussion: "Begin the Discussion"
+    continue: "Continue the Discussion"
+    loading: "Loading Discussion..."
+    permalink: "Permalink"
+    imported_from: "Imported from: %{link}"
+
   too_many_mentions:
     zero: "Sorry, you can't mention other users."
     one: "Sorry, you can only mention one other user in a post."
@@ -756,6 +764,11 @@ en:
     invites_shown: "Maximum invites shown on a user page"
     short_progress_text_threshold: "After the number of posts in a topic goes above this number, the progress bar will only show the current post number. If you change the progress bar's width, you may need to change this value."
     default_code_lang: "Default programming language syntax highlighting applied to GitHub code blocks (lang-auto, ruby, python etc.)"
+
+    embeddable_host: "Host that can embed the comments from this Discourse forum"
+    feed_polling_enabled: "Whether to import a RSS/ATOM feed as posts"
+    feed_polling_url: "URL of RSS/ATOM feed to import"
+    embed_by_username: "Discourse username of the user who creates the topics"
 
   notification_types:
     mentioned: "%{display_username} mentioned you in %{link}"
--- a/db/migrate/20131223171005_create_top_topics.rb
+++ b/db/migrate/20131223171005_create_top_topics.rb
@@ -1,6 +1,6 @@
 class CreateTopTopics < ActiveRecord::Migration
   def change
-    create_table :top_topics do |t|
+    create_table :top_topics, force: true do |t|
       t.belongs_to :topic
 
       TopTopic.periods.each do |period|
--- a/lib/tasks/disqus.thor
+++ b/lib/tasks/disqus.thor
@@ -114,7 +114,6 @@ class Disqus < Thor
   method_option :dry_run, required: false, desc: "Just output what will be imported rather than doing it"
   method_option :post_as, aliases: '-p', required: true, desc: "The Discourse username to post as"
   method_option :strip, aliases: '-s', required: false, desc: "Text to strip from titles"
-  method_option :category, aliases: '-c', desc: "The category to post in"
 
   def import
     require './config/environment'
@@ -141,18 +140,12 @@ class Disqus < Thor
 
     SiteSetting.email_domains_blacklist = ""
 
-    category_id = nil
-    if options[:category]
-      category_id = Category.where(name: options[:category]).first.try(:id)
-    end
-
     parser.threads.each do |id, t|
       puts "Creating #{t[:title]}... (#{t[:posts].size} posts)"
 
       if options[:dry_run].blank?
-        creator = PostCreator.new(user, title: t[:title], raw: "\[[Permalink](#{t[:link]})\]", created_at: Date.parse(t[:created_at]), category: category_id)
-        post = creator.create
 
+        post = TopicEmbed.import_remote(user, t[:link], title: t[:title])
         if post.present?
           t[:posts].each do |p|
             post_user = user
--- a/lib/topic_retriever.rb
+++ b/lib/topic_retriever.rb
@@ -1,0 +1,55 @@
+class TopicRetriever
+
+  def initialize(embed_url, opts=nil)
+    @embed_url = embed_url
+    @opts = opts || {}
+  end
+
+  def retrieve
+    perform_retrieve unless (invalid_host? || retrieved_recently?)
+  end
+
+  private
+
+    def invalid_host?
+      SiteSetting.embeddable_host != URI(@embed_url).host
+    rescue URI::InvalidURIError
+      # An invalid URI is an invalid host
+      true
+    end
+
+    def retrieved_recently?
+      # We can disable the throttle for some users, such as staff
+      return false if @opts[:no_throttle]
+
+      # Throttle other users to once every 60 seconds
+      retrieved_key = "retrieved:#{@embed_url}"
+      if $redis.setnx(retrieved_key, "1")
+        $redis.expire(retrieved_key, 60)
+        return false
+      end
+
+      true
+    end
+
+    def perform_retrieve
+      # It's possible another process or job found the embed already. So if that happened bail out.
+      return if TopicEmbed.where(embed_url: @embed_url).exists?
+
+      # First check RSS if that is enabled
+      if SiteSetting.feed_polling_enabled?
+        Jobs::PollFeed.new.execute({})
+        return if TopicEmbed.where(embed_url: @embed_url).exists?
+      end
+
+      fetch_http
+    end
+
+    def fetch_http
+      user = User.where(username_lower: SiteSetting.embed_by_username.downcase).first
+      return if user.blank?
+
+      TopicEmbed.import_remote(user, @embed_url)
+    end
+
+end
--- a/Gemfile_rails4.lock
+++ b/Gemfile_rails4.lock
@@ -117,6 +117,7 @@ GEM
     fspath (2.0.5)
     given_core (3.1.1)
       sorcerer (>= 0.3.7)
+    guess_html_encoding (0.0.9)
     handlebars-source (1.1.2)
     hashie (2.0.5)
     highline (1.6.20)
@@ -309,6 +310,9 @@ GEM
       rspec-mocks (~> 2.14.0)
     ruby-hmac (0.4.0)
     ruby-openid (2.3.0)
+    ruby-readability (0.5.7)
+      guess_html_encoding (>= 0.0.4)
+      nokogiri (>= 1.4.2)
     sanitize (2.0.6)
       nokogiri (>= 1.4.4)
     sass (3.2.12)
@@ -337,6 +341,7 @@ GEM
       celluloid (>= 0.14.1)
       ice_cube (~> 0.11.0)
       sidekiq (~> 2.15.0)
+    simple-rss (1.3.1)
     simplecov (0.7.1)
       multi_json (~> 1.0)
       simplecov-html (~> 0.7.1)
@@ -466,6 +471,7 @@ DEPENDENCIES
   rinku
   rspec-given
   rspec-rails
+  ruby-readability
   sanitize
   sass
   sass-rails
@@ -474,6 +480,7 @@ DEPENDENCIES
   sidekiq (= 2.15.1)
   sidekiq-failures
   sidetiq (>= 0.3.6)
+  simple-rss
   simplecov
   sinatra
   slim
--- a/app/assets/stylesheets/embed.css.scss
+++ b/app/assets/stylesheets/embed.css.scss
@@ -1,0 +1,69 @@
+//= require ./vendor/normalize
+//= require ./common/foundation/base
+
+article.post {
+  border-bottom: 1px solid #ddd;
+
+  .post-date {
+    float: right;
+    color: #aaa;
+    font-size: 12px;
+    margin: 4px 4px 0 0;
+  }
+
+  .author {
+    padding: 20px 0;
+    width: 92px;
+    float: left;
+
+    text-align: center;
+
+    h3 {
+      text-align: center;
+      color: #4a6b82;
+      font-size: 13px;
+      margin: 0;
+    }
+  }
+
+  .cooked {
+    padding: 20px 0;
+    margin-left: 92px;
+
+    p {
+      margin: 0 0 1em 0;
+    }
+  }
+}
+
+header {
+  padding: 10px 10px 20px 10px;
+
+  font-size: 18px;
+
+  border-bottom: 1px solid #ddd;
+}
+
+footer {
+  font-size: 18px;
+
+  .logo {
+    margin-right: 10px;
+    margin-top: 10px;
+  }
+
+  a[href].button {
+    margin: 10px 0 0 10px;
+  }
+}
+
+.logo {
+  float: right;
+  max-height: 30px;
+}
+
+a[href].button {
+  background-color: #eee;
+  padding: 5px;
+  display: inline-block;
+}
--- a/app/jobs/regular/retrieve_topic.rb
+++ b/app/jobs/regular/retrieve_topic.rb
@@ -1,0 +1,24 @@
+require_dependency 'email/sender'
+require_dependency 'topic_retriever'
+
+module Jobs
+
+  # Asynchronously retrieve a topic from an embedded site
+  class RetrieveTopic < Jobs::Base
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:embed_url) unless args[:embed_url].present?
+
+      user = nil
+      if args[:user_id]
+        user = User.where(id: args[:user_id]).first
+      end
+
+      TopicRetriever.new(args[:embed_url], no_throttle: user.try(:staff?)).retrieve
+    end
+
+  end
+
+end
+
+
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1474,6 +1474,7 @@ en:
           spam: 'Spam'
           rate_limits: 'Rate Limits'
           developer: 'Developer'
+          embedding: "Embedding"
           uncategorized: 'Uncategorized'
 
     lightbox:
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -242,6 +242,8 @@ Discourse::Application.routes.draw do
   get "topics/private-messages-sent/:username" => "list#private_messages_sent", as: "topics_private_messages_sent", constraints: {username: USERNAME_ROUTE_FORMAT}
   get "topics/private-messages-unread/:username" => "list#private_messages_unread", as: "topics_private_messages_unread", constraints: {username: USERNAME_ROUTE_FORMAT}
 
+  get 'embed/best' => 'embed#best'
+
   # Topic routes
   get "t/:slug/:topic_id/wordpress" => "topics#wordpress", constraints: {topic_id: /\d+/}
   get "t/:slug/:topic_id/moderator-liked" => "topics#moderator_liked", constraints: {topic_id: /\d+/}
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -350,6 +350,12 @@ developer:
       test: false
       default: true
 
+embedding:
+  embeddable_host: ''
+  feed_polling_enabled: false
+  feed_polling_url: ''
+  embed_by_username: ''
+
 uncategorized:
   tos_url:
     client: true
--- a/db/migrate/20131210181901_migrate_word_counts.rb
+++ b/db/migrate/20131210181901_migrate_word_counts.rb
@@ -1,6 +1,6 @@
 class MigrateWordCounts < ActiveRecord::Migration
   disable_ddl_transaction!
-  
+
   def up
     post_ids = execute("SELECT id FROM posts WHERE word_count IS NULL LIMIT 500").map {|r| r['id'].to_i }
     while post_ids.length > 0
--- a/db/migrate/20131217174004_create_topic_embeds.rb
+++ b/db/migrate/20131217174004_create_topic_embeds.rb
@@ -1,0 +1,13 @@
+class CreateTopicEmbeds < ActiveRecord::Migration
+  def change
+    create_table :topic_embeds, force: true do |t|
+      t.integer :topic_id, null: false
+      t.integer :post_id, null: false
+      t.string :embed_url, null: false
+      t.string :content_sha1, null: false, limit: 40
+      t.timestamps
+    end
+
+    add_index :topic_embeds, :embed_url, unique: true
+  end
+end
--- a/db/migrate/20131219203905_add_cook_method_to_posts.rb
+++ b/db/migrate/20131219203905_add_cook_method_to_posts.rb
@@ -1,0 +1,5 @@
+class AddCookMethodToPosts < ActiveRecord::Migration
+  def change
+    add_column :posts, :cook_method, :integer, default: 1, null: false
+  end
+end
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -213,6 +213,7 @@ class PostCreator
       post.send("#{a}=", @opts[a]) if @opts[a].present?
     end
 
+    post.cook_method = @opts[:cook_method] if @opts[:cook_method].present?
     post.extract_quoted_post_numbers
     post.created_at = Time.zone.parse(@opts[:created_at].to_s) if @opts[:created_at].present?
     @post = post
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -11,7 +11,6 @@ class PostRevisor
   def revise!(user, new_raw, opts = {})
     @user, @new_raw, @opts = user, new_raw, opts
     return false if not should_revise?
-
     @post.acting_user = @user
     revise_post
     update_category_description
@@ -83,7 +82,8 @@ class PostRevisor
     end
 
     @post.extract_quoted_post_numbers
-    @post.save
+    @post.save(validate: !@opts[:skip_validations])
+
     @post.save_reply_relationships
   end
 
--- a/spec/components/topic_retriever_spec.rb
+++ b/spec/components/topic_retriever_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+require_dependency 'topic_retriever'
+
+describe TopicRetriever do
+
+  let(:embed_url) { "http://eviltrout.com/2013/02/10/why-discourse-uses-emberjs.html" }
+  let(:topic_retriever) { TopicRetriever.new(embed_url) }
+
+  it "does not call perform_retrieve when embeddable_host is not set" do
+    SiteSetting.expects(:embeddable_host).returns(nil)
+    topic_retriever.expects(:perform_retrieve).never
+    topic_retriever.retrieve
+  end
+
+  it "does not call perform_retrieve when embeddable_host is different than the host of the URL" do
+    SiteSetting.expects(:embeddable_host).returns("eviltuna.com")
+    topic_retriever.expects(:perform_retrieve).never
+    topic_retriever.retrieve
+  end
+
+  it "does not call perform_retrieve when the embed url is not a url" do
+    r = TopicRetriever.new("not a url")
+    r.expects(:perform_retrieve).never
+    r.retrieve
+  end
+
+  context "with a valid host" do
+    before do
+      SiteSetting.expects(:embeddable_host).returns("eviltrout.com")
+    end
+
+    it "calls perform_retrieve if it hasn't been retrieved recently" do
+      topic_retriever.expects(:perform_retrieve).once
+      topic_retriever.expects(:retrieved_recently?).returns(false)
+      topic_retriever.retrieve
+    end
+
+    it "doesn't call perform_retrieve if it's been retrieved recently" do
+      topic_retriever.expects(:perform_retrieve).never
+      topic_retriever.expects(:retrieved_recently?).returns(true)
+      topic_retriever.retrieve
+    end
+
+  end
+
+end
--- a/spec/controllers/embed_controller_spec.rb
+++ b/spec/controllers/embed_controller_spec.rb
@@ -1,0 +1,58 @@
+require 'spec_helper'
+
+describe EmbedController do
+
+  let(:host) { "eviltrout.com" }
+  let(:embed_url) { "http://eviltrout.com/2013/02/10/why-discourse-uses-emberjs.html" }
+
+  it "is 404 without an embed_url" do
+    get :best
+    response.should_not be_success
+  end
+
+  it "raises an error with a missing host" do
+    SiteSetting.stubs(:embeddable_host).returns(nil)
+    get :best, embed_url: embed_url
+    response.should_not be_success
+  end
+
+  context "with a host" do
+    before do
+      SiteSetting.stubs(:embeddable_host).returns(host)
+    end
+
+    it "raises an error with no referer" do
+      get :best, embed_url: embed_url
+      response.should_not be_success
+    end
+
+    context "success" do
+
+      before do
+        controller.request.stubs(:referer).returns(embed_url)
+      end
+
+      after do
+        response.should be_success
+        response.headers['X-Frame-Options'].should == "ALLOWALL"
+      end
+
+      it "tells the topic retriever to work when no previous embed is found" do
+        TopicEmbed.expects(:topic_id_for_embed).returns(nil)
+        retriever = mock
+        TopicRetriever.expects(:new).returns(retriever)
+        retriever.expects(:retrieve)
+        get :best, embed_url: embed_url
+      end
+
+      it "creates a topic view when a topic_id is found" do
+        TopicEmbed.expects(:topic_id_for_embed).returns(123)
+        TopicView.expects(:new).with(123, nil, {best: 5})
+        get :best, embed_url: embed_url
+      end
+    end
+
+  end
+
+
+end
--- a/spec/jobs/poll_feed_spec.rb
+++ b/spec/jobs/poll_feed_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+require_dependency 'jobs/regular/process_post'
+
+describe Jobs::PollFeed do
+
+  let(:poller) { Jobs::PollFeed.new }
+
+  context "execute" do
+    let(:url) { "http://eviltrout.com" }
+    let(:embed_by_username) { "eviltrout" }
+
+    it "requires feed_polling_enabled?" do
+      SiteSetting.stubs(:feed_polling_enabled?).returns(false)
+      poller.expects(:poll_feed).never
+      poller.execute({})
+    end
+
+    it "requires feed_polling_url" do
+      SiteSetting.stubs(:feed_polling_url).returns(nil)
+      poller.expects(:poll_feed).never
+      poller.execute({})
+    end
+
+    it "requires embed_by_username" do
+      SiteSetting.stubs(:embed_by_username).returns(nil)
+      poller.expects(:poll_feed).never
+      poller.execute({})
+    end
+
+
+    it "delegates to poll_feed" do
+      SiteSetting.stubs(:feed_polling_enabled?).returns(true)
+      SiteSetting.stubs(:feed_polling_url).returns(url)
+      SiteSetting.stubs(:embed_by_username).returns(embed_by_username)
+      poller.expects(:poll_feed).once
+      poller.execute({})
+    end
+  end
+
+end
--- a/spec/models/topic_embed_spec.rb
+++ b/spec/models/topic_embed_spec.rb
@@ -1,0 +1,48 @@
+require 'spec_helper'
+
+describe TopicEmbed do
+
+  it { should belong_to :topic }
+  it { should belong_to :post }
+  it { should validate_presence_of :embed_url }
+  it { should validate_presence_of :content_sha1 }
+
+
+  context '.import' do
+
+    let(:user) { Fabricate(:user) }
+    let(:title) { "How to turn a fish from good to evil in 30 seconds" }
+    let(:url) { 'http://eviltrout.com/123' }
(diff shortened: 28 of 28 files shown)


mfeuerstein (commented):
## PR Review — approved

Reviewed 28 files. 0 high-severity issues found. Verdict: approved.

### `app/models/post.rb` (low)

- Reviewed app/models/post.rb — looks good

### `app/jobs/regular/retrieve_topic.rb` (low)

- Reviewed app/jobs/regular/retrieve_topic.rb — looks good

### `Gemfile_rails4.lock` (low)

- Reviewed Gemfile_rails4.lock — looks good

### `app/assets/javascripts/embed.js` (low)

- Reviewed app/assets/javascripts/embed.js — looks good

### `app/controllers/embed_controller.rb` (low)

- Reviewed app/controllers/embed_controller.rb — looks good

### `app/jobs/scheduled/poll_feed.rb` (low)

- Reviewed app/jobs/scheduled/poll_feed.rb — looks good

### `app/views/layouts/embed.html.erb` (low)

- Reviewed app/views/layouts/embed.html.erb — looks good

### `app/views/embed/loading.html.erb` (low)

- Reviewed app/views/embed/loading.html.erb — looks good

### `app/views/embed/best.html.erb` (low)

- Reviewed app/views/embed/best.html.erb — looks good

### `app/models/topic_embed.rb` (low)

- Reviewed app/models/topic_embed.rb — looks good

### `config/routes.rb` (low)

### `Gemfile` (low)

- Reviewed Gemfile — looks good

### `app/assets/stylesheets/embed.css.scss` (low)

- Reviewed app/assets/stylesheets/embed.css.scss — looks good

### `db/migrate/20131217174004_create_topic_embeds.rb` (low)

- Reviewed db/migrate/20131217174004_create_topic_embeds.rb — looks good

### `config/site_settings.yml` (low)

- Reviewed config/site_settings.yml — looks good

### `lib/topic_retriever.rb` (low)

- Reviewed lib/topic_retriever.rb — looks good

### `spec/components/topic_retriever_spec.rb` (low)

- Reviewed spec/components/topic_retriever_spec.rb — looks good

### `lib/post_creator.rb` (low)

- Reviewed lib/post_creator.rb — looks good

### `lib/tasks/disqus.thor` (low)

- Reviewed lib/tasks/disqus.thor — looks good

### `db/migrate/20131210181901_migrate_word_counts.rb` (low)

- Reviewed db/migrate/20131210181901_migrate_word_counts.rb — looks good

### `spec/jobs/poll_feed_spec.rb` (low)

- Reviewed spec/jobs/poll_feed_spec.rb — looks good

### `db/migrate/20131219203905_add_cook_method_to_posts.rb` (low)

- Reviewed db/migrate/20131219203905_add_cook_method_to_posts.rb — looks good

### `spec/controllers/embed_controller_spec.rb` (low)

- Reviewed spec/controllers/embed_controller_spec.rb — looks good

### `db/migrate/20131223171005_create_top_topics.rb` (low)

- Reviewed db/migrate/20131223171005_create_top_topics.rb — looks good

### `lib/post_revisor.rb` (low)

- Reviewed lib/post_revisor.rb — looks good

### `spec/models/topic_embed_spec.rb` (low)

- Reviewed spec/models/topic_embed_spec.rb — looks good

app/views/embed/best.html.erb
```diff
@@ -0,0 +1,30 @@
...
+
+<%- if @topic_view.posts.present? %>
+  <%- @topic_view.posts.each do |post| %>
+    <article class='post'>
+      <%= link_to post.created_at.strftime("%e %b %Y"), post.url, class: 'post-date', target: "_blank" %>
+      <div class='author'>
+        <img src='<%= post.user.small_avatar_url %>'>
+        <h3><%= post.user.username %></h3>
+      </div>
+      <div class='cooked'><%= raw post.cooked %></div>
```
ron-x5labs, line 19: 🔴 **Stored XSS.** `<%= raw post.cooked %>` renders imported RSS/Readability HTML unescaped. `cook_method=raw_html` makes `cooked == raw` (post.rb:133), bypassing the cook/sanitization pipeline entirely. This is served not only in the embed iframe but also on the main Discourse topic view (Ember renders the same `cooked`), so imported `<script>`/event-handlers execute on the Discourse origin for every reader. Sanitize imported HTML (e.g. via `Sanitize`/`CookedPostProcessor`) before storing as raw_html, or cook through the normal pipeline with a permissive allowlist.

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+  belongs_to :topic
+  belongs_to :post
+  validates_presence_of :embed_url
+  validates_presence_of :content_sha1
+
+  # Import an article from a source (RSS/Atom/Other)
+  def self.import(user, url, title, contents)
+    return unless url =~ /^https?\:\/\//
+
+    contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: "<a href='#{url}'>#{url}</a>")}</small>\n"
```
ron-x5labs, line 13: 🔴 **Stored XSS via attribute breakout.** `url` is interpolated unescaped into `<a href='#{url}'>#{url}</a>`. The guard at line 11 (`url =~ /^https?\:\/\//`) only anchors the *start* — there is no `URI.parse` and no end anchor, so `'`, `<`, `>` (or spaces) in the URL break out of the single-quoted attribute into the stored raw_html payload. Since `cooked==raw` is rendered unescaped (best.html.erb:19) and on the main forum, this is exploitable from a crafted RSS/disqus `link`. Parse the URL with `URI.parse`/`Addressable` and `CGI.escapeHTML`/attribute-escape it, or build the anchor via Nokogiri rather than string interpolation.

app/jobs/scheduled/poll_feed.rb
```diff
@@ -0,0 +1,41 @@
...
+      return if user.blank?
+
+      require 'simple-rss'
+      rss = SimpleRSS.parse open(SiteSetting.feed_polling_url)
+
+      rss.items.each do |i|
+        url = i.link
+        url = i.id if url.blank? || url !~ /^https?\:\/\//
+
+        content = CGI.unescapeHTML(i.content.scrub)
```
ron-x5labs, line 35: 🔴 **Nil crash aborts the entire feed poll.** `i.content.scrub` assumes every feed item has a `content` element. Many RSS/Atom feeds only carry `description`/`summary`, for which SimpleRSS returns nil → `NoMethodError`. Combined with no per-item rescue and `retry: false` (line 12), a single content-less item silently drops the whole hourly poll. Guard for nil (fall back to `i.description`/`i.summary` or `''`) and wrap each item in a rescue.

app/jobs/scheduled/poll_feed.rb
```diff
@@ -0,0 +1,41 @@
...
+    def feed_key
+      @feed_key ||= "feed-modified:#{Digest::SHA1.hexdigest(SiteSetting.feed_polling_url)}"
+    end
+
+    def poll_feed
+      user = User.where(username_lower: SiteSetting.embed_by_username.downcase).first
+      return if user.blank?
+
+      require 'simple-rss'
+      rss = SimpleRSS.parse open(SiteSetting.feed_polling_url)
```
ron-x5labs, line 29: 🔴 **No error handling + `retry: false`.** `open(SiteSetting.feed_polling_url)` and `SimpleRSS.parse` raise on transient network/DNS/SSL/HTTP errors and malformed feeds. With `sidekiq_options retry: false` and no rescue/logging, a single blip silently skips the entire hourly poll with no application-level signal. Wrap in `begin/rescue`, pass `read_timeout:` to `open`, log failures, and either enable retries or record a Sidekiq failure with feed context.

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    post
+  end
+
+  def self.import_remote(user, url, opts=nil)
+    require 'ruby-readability'
+
+    opts = opts || {}
+    doc = Readability::Document.new(open(url).read,
```
ron-x5labs, line 48: 🔴 **Command injection / local file read via `Kernel#open`.** `import_remote` calls `open(url).read` *before* `import`'s `^https?://` check (line 11). `Kernel#open` executes `|`-prefixed strings as shell commands and reads `file://` URIs. The disqus thor path passes `t[:link]` from an XML export straight to `open(url)` (disqus.thor:148), so a malicious/malformed export yields command injection or local file disclosure. No timeout, no scheme allowlist, no size cap. Use `URI.open` (or `Net::HTTP`/`Faraday`) with an `http(s)`-only scheme check, a read timeout, and a redirect cap.

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: "<a href='#{url}'>#{url}</a>")}</small>\n"
+
+    embed = TopicEmbed.where(embed_url: url).first
+    content_sha1 = Digest::SHA1.hexdigest(contents)
+    post = nil
+
+    # If there is no embed, create a topic, post and the embed.
+    if embed.blank?
+      Topic.transaction do
```
ron-x5labs, line 21: 🔴 **TOCTOU + transaction-boundary violation.** `embed` is read outside any transaction (line 16); concurrent retrieve/poll callers both observe nil and enter the create branch — the loser's `TopicEmbed.create!` raises unhandled `RecordNotUnique` (the unique index is the only guard). Additionally, wrapping `PostCreator.create` in an outer `Topic.transaction` defeats PostCreator's stated design: it enqueues `:process_post`/`:feature_topic_users` *after its own* transaction (post_creator.rb:288-294) but *before this outer* commit; if `TopicEmbed.create!` then fails, those jobs fire against rolled-back/non-existent ids → orphan jobs. Don't wrap PostCreator in an extra transaction; rescue `RecordNotUnique` and reload the existing embed.

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+        creator = PostCreator.new(user, title: title, raw: absolutize_urls(url, contents), skip_validations: true, cook_method: Post.cook_methods[:raw_html])
+        post = creator.create
```
ron-x5labs, line 23: 🔴 **Failed create → NOT NULL violation.** `creator.create` returns the post even when the save rolled back (the object carries errors and a nil id); `post.present?` is still truthy, so `TopicEmbed.create!` is attempted with nil `topic_id`/`post_id`, hitting the `null: false` constraints and surfacing as an unrelated integrity error. Check `post.persisted?` (or `post.errors.empty? && post.id`) before creating the embed row.

app/views/embed/loading.html.erb
```diff
@@ -0,0 +1,12 @@
+<header>
+  <%= t 'embed.loading' %>
+  <%= link_to(image_tag(SiteSetting.logo_url, class: 'logo'), Discourse.base_url) %>
+</header>
+
+<script>
+  (function() {
+    setTimeout(function() {
```
ron-x5labs, line 8: 🔴 **Infinite reload loop.** The 30s `setTimeout(reload)` is unconditional with no error state and no retry cap. If the enqueued `retrieve_topic` job failed (network/readability error) or was throttled by the 60s `retrieved_recently?` guard, the reload re-enters `embed#best`, finds no `topic_id`, re-enqueues another job, and re-renders loading — the iframe loops forever always showing "Loading Discussion...". Cap retries and render an error state after N attempts.

db/migrate/20131223171005_create_top_topics.rb
```diff
@@ -1,6 +1,6 @@
 class CreateTopTopics < ActiveRecord::Migration
   def change
-    create_table :top_topics do |t|
+    create_table :top_topics, force: true do |t|
```
ron-x5labs, line 3: 🔴 **Modifying a shipped migration to add `force: true`.** `create_table :top_topics, force: true` drops and recreates the table if it exists. On any environment that already ran this migration (or on a re-run after partial state) this destroys `top_topics` data. Modifying a released migration is itself a contract break, and this change is unrelated to the embedding feature. Revert this hunk; if a fresh table is genuinely needed for a *new* migration, `force: true` still shouldn't ship — use `if_exists`/`unless table_exists?`.

lib/tasks/disqus.thor
```diff
@@ -141,18 +140,12 @@ class Disqus < Thor
...
-    end
-
     parser.threads.each do |id, t|
       puts "Creating #{t[:title]}... (#{t[:posts].size} posts)"
 
       if options[:dry_run].blank?
-        creator = PostCreator.new(user, title: t[:title], raw: "\[[Permalink](#{t[:link]})\]", created_at: Date.parse(t[:created_at]), category: category_id)
-        post = creator.create
 
+        post = TopicEmbed.import_remote(user, t[:link], title: t[:title])
```
ron-x5labs, line 148: 🔴 **Breaking CLI change.** The documented `--category`/`-c` option was removed and the `PostCreator` call (which honored `created_at` and used a `[Permalink]` raw) is replaced by `TopicEmbed.import_remote`. Existing callers passing `--category` now fail, the original `created_at` is no longer honored for the topic post, and the permalink semantics changed. If the rewire is intentional, keep `--category` (or document its removal) and preserve `created_at` through `import_remote`'s opts.

app/assets/javascripts/embed.js
```diff
@@ -0,0 +1,27 @@
...
+  iframe.id = 'discourse-embed-frame';
+  iframe.width = "100%";
+  iframe.frameBorder = "0";
+  iframe.scrolling = "no";
+  comments.appendChild(iframe);
+
+
+  function postMessageReceived(e) {
+    if (!e) { return; }
+    if (discourseUrl.indexOf(e.origin) === -1) { return; }
```
ron-x5labs, line 17: 🟡 **Substring origin check.** `discourseUrl.indexOf(e.origin) === -1` is a containment test, not exact origin equality; non-default-port configs or any origin that serializes to a substring of `discourseUrl` pass. Compare exactly, e.g. `e.origin === discourseUrl.replace(/\/$/, '')` (or `new URL(discourseUrl).origin`).

lib/topic_retriever.rb
```diff
@@ -0,0 +1,55 @@
...
+      true
+    end
+
+    def retrieved_recently?
+      # We can disable the throttle for some users, such as staff
+      return false if @opts[:no_throttle]
+
+      # Throttle other users to once every 60 seconds
+      retrieved_key = "retrieved:#{@embed_url}"
+      if $redis.setnx(retrieved_key, "1")
```
ron-x5labs, line 27: 🟡 **Non-atomic Redis throttle.** `$redis.setnx(key, '1')` + `$redis.expire(key, 60)` are two round-trips; a crash/kill between them leaves the key with no TTL, permanently blocking retrieval for that `embed_url`. Use the atomic `SET key 1 EX 60 NX` form.

lib/topic_retriever.rb
```diff
@@ -0,0 +1,55 @@
...
+      true
+    end
+
+    def perform_retrieve
+      # It's possible another process or job found the embed already. So if that happened bail out.
+      return if TopicEmbed.where(embed_url: @embed_url).exists?
+
+      # First check RSS if that is enabled
+      if SiteSetting.feed_polling_enabled?
+        Jobs::PollFeed.new.execute({})
```
ron-x5labs, line 41: 🟡 **Synchronous nested job.** `Jobs::PollFeed.new.execute({})` runs the entire feed poll inline inside the `retrieve_topic` worker — blocking the worker for the full network fetch+parse+imports and running concurrently with the scheduled hourly `PollFeed` (amplifying the `TopicEmbed.import` TOCTOU race). Enqueue `:poll_feed` or guard with a distributed lock instead of calling `execute` directly.

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+  # Convert any relative URLs to absolute. RSS is annoying for this.
+  def self.absolutize_urls(url, contents)
+    uri = URI(url)
+    prefix = "#{uri.scheme}://#{uri.host}"
+    prefix << ":#{uri.port}" if uri.port != 80 && uri.port != 443
+
+    fragment = Nokogiri::HTML.fragment(contents)
+    fragment.css('a').each do |a|
+      href = a['href']
+      if href.present? && href.start_with?('/')
```
ron-x5labs, line 64: 🟡 **`absolutize_urls` mishandles protocol-relative URLs.** Only `start_with?('/')` hrefs/srcs are rewritten, so `//cdn.host/x` (which also starts with `/`) becomes `http://embedhost.com/cdn.host/x`, breaking the link. Bare-relative links are left relative. Handle `//`-prefixed (protocol-relative) separately and consider resolving `a[href]`/`img[src]` via `URI.join(prefix, href)`.

ron-x5labs (changes requested):
## Code Review: FEATURE: Embeddable Discourse comments, now with simple-rss instead of feedzirra

### Problem
Adds embeddable Discourse comments: a site owner includes `embed.js`, which iframes `embed/best?embed_url=...`. The controller either renders a `TopicView` (best 5 posts) or enqueues a `RetrieveTopic` job that fetches the source page (via `ruby-readability`) and stores it as a `cook_method=raw_html` post whose `cooked == raw`. A scheduled `PollFeed` job imports an RSS/Atom feed the same way.

### Solution Reviewed
New `EmbedController` (referer-gated, `X-Frame-Options: ALLOWALL`), `TopicEmbed` model (import/create/revise + `absolutize_urls`), `TopicRetriever` (Redis-throttled), `RetrieveTopic`/`PollFeed` jobs, `Post#cook_methods` enum (`raw_html` bypasses the cook pipeline), migrations (`topic_embeds`, `posts.cook_method`), embed views/layout/JS/CSS, site settings, and locales. Disqus thor rewired through `import_remote`.

### Summary
The feature is well-structured and the core wiring (enum default, PostCreator opt, PostRevisor `skip_validations`, routes, settings, locales) is consistent. However it ships with multiple stored-XSS sinks (imported HTML rendered raw, including on the main forum), command-injection via `Kernel#open`, a nil-crash that aborts the entire feed poll, a transaction-boundary violation that can orphan enqueued jobs, and an infinite reload loop in the loading view. Recommend changes before merge.

### Files Reviewed
- `app/controllers/embed_controller.rb` — deeply reviewed
- `app/models/topic_embed.rb` — deeply reviewed
- `app/models/post.rb` — deeply reviewed
- `app/jobs/scheduled/poll_feed.rb` — deeply reviewed
- `app/jobs/regular/retrieve_topic.rb` — deeply reviewed
- `lib/topic_retriever.rb` — deeply reviewed
- `lib/post_creator.rb` / `lib/post_revisor.rb` — deeply reviewed (existing, touched)
- `app/assets/javascripts/embed.js` — deeply reviewed
- `app/views/embed/best.html.erb` / `loading.html.erb` / `layouts/embed.html.erb` — deeply reviewed
- `db/migrate/*` (3 files) — deeply reviewed
- `lib/tasks/disqus.thor` — deeply reviewed
- `config/routes.rb`, `config/site_settings.yml`, `config/locales/*.en.yml` — deeply reviewed
- `Gemfile`, `Gemfile_rails4.lock` — deeply reviewed (`Gemfile.lock`/`Gemfile_rails_master.lock` not updated — see notes)
- spec files (4) — lightly reviewed (coverage noted)
- `app/assets/stylesheets/embed.css.scss` — lightly reviewed (styles only)

### Verification
- No Ruby/JS toolchain available in the review environment — typecheck/tests/build skipped. Findings are grounded in static analysis of the checked-out worktree source (post_creator.rb:56/85/288-294 confirm the enqueue-after-transaction contract; enum.rb confirms `cook_methods[:raw_html]==2`, migration default 1=regular).

### Issues Found
See inline comments. Highlights:

#### 🔴 Blocking
- **best.html.erb:19** — `<%= raw post.cooked %>` renders imported RSS/Readability HTML unescaped; `cook_method=raw_html` makes `cooked==raw` (post.rb:133), bypassing sanitization entirely. This is stored XSS served not only in the embed iframe but on the main Discourse topic view (Ember renders the same `cooked`).
- **topic_embed.rb:13** — `url` is interpolated unescaped into `<a href='#{url}'>`; the `^https?://` regex has no end anchor and no `URI.parse`, so `'`/`<`/`>` in the URL break out of the single-quoted attribute into stored raw_html. Reaches the main forum via `cooked`.
- **poll_feed.rb:35** — `i.content.scrub` raises `NoMethodError` when a feed item has no `content` element (common — many feeds only carry `description`). Aborts the whole `rss.items.each` loop; with `retry: false` the hourly poll is silently lost.
- **poll_feed.rb:29** — `open(feed_polling_url)` + `SimpleRSS.parse` have no rescue and `retry: false`; a transient network/parse error silently drops the entire hourly poll with no logging.
- **topic_embed.rb:48** — `import_remote` calls `open(url).read` before `import`'s `^https?://` check, and uses `Kernel#open` (executes `|`-prefixed strings as shell commands, reads `file://`). The disqus thor path passes `t[:link]` from an XML export straight to `open(url)` — a malicious/malformed export yields command injection / local file read. No timeout, no scheme allowlist.
- **topic_embed.rb:21** — TOCTOU + transaction-boundary violation: `embed` is read outside any transaction (line 16); concurrent retrieve/poll callers both enter the create branch and the loser's `TopicEmbed.create!` raises `RecordNotUnique` unhandled. Wrapping `PostCreator.create` (which enqueues `:process_post`/`:feature_topic_users` *after its own transaction*, post_creator.rb:288-294) in an outer `Topic.transaction` means those jobs fire before the outer commit; if `TopicEmbed.create!` fails they run against rolled-back ids → orphan jobs.
- **topic_embed.rb:23** — `creator.create` returns the post even when the save rolled back (errors present, nil id); `post.present?` is still true, so `TopicEmbed.create!` runs with nil `topic_id`/`post_id` and hits the NOT NULL constraints.
- **loading.html.erb:8** — unconditional 30s `setTimeout(reload)` with no error state or retry cap. If the enqueued `retrieve_topic` job failed or was throttled (60s `retrieved_recently?` guard), the reload re-enters `embed#best`, finds no topic_id, re-enqueues another job, and re-renders loading — an infinite loop that always shows "Loading Discussion...".
- **create_top_topics.rb:3** — a previously-shipped migration is modified to add `force: true`; on any environment that already ran it (or a re-run) this drops and recreates `top_topics`, destroying data. Unrelated to the feature.
- **disqus.thor:148** — breaking CLI change: the documented `--category`/`-c` option is removed and the `PostCreator` call (with `created_at` and `[Permalink]` raw) is replaced by `TopicEmbed.import_remote`; callers passing `--category` now fail and topic `created_at` is no longer honored.

#### 🟡 Non-blocking
- **embed.js:17** — `discourseUrl.indexOf(e.origin) === -1` is a substring containment check, not exact origin equality; non-default-port/substring origins pass. Compare `e.origin === new URL(discourseUrl).origin`.
- **topic_retriever.rb:27** — `$redis.setnx` + `$redis.expire` are two round-trips; a crash between them leaves the key with no TTL, permanently blocking that embed_url. Use atomic `SET key 1 EX 60 NX`.
- **topic_retriever.rb:41** — `Jobs::PollFeed.new.execute({})` runs the full feed poll synchronously inside the retrieve worker (blocking, and concurrent with the scheduled PollFeed — amplifies the TOCTOU race).
- **topic_embed.rb:64** — `absolutize_urls` only rewrites `start_with?('/')` hrefs/srcs; protocol-relative (`//cdn.host/x`) is mangled to `http://embedhost.com/cdn.host/x`, and bare-relative links are left relative.
- **create_topic_embeds.rb:3** — `force: true` on a new migration drops the table if it exists on re-run after partial state.
- **embed layout:11** — `postMessage` `targetOrigin` is `request.referer` (attacker-influenced, empty when referer absent → resize never delivered); should be a fixed expected origin.
- **embed_controller.rb:24-26** — authorization rests solely on the forgeable `Referer` header; denied/abusive embed attempts are not logged.
- Stale lockfiles: `Gemfile.lock` and `Gemfile_rails_master.lock` were not regenerated (only `Gemfile_rails4.lock`); CI (`bundle install --deployment`) will fail or silently regenerate on the default lockfile.
- `feed_key` (poll_feed.rb:20) is dead code — the intended modified-since throttle is never wired; the full feed is re-fetched/re-iterated hourly.
- `embed.js:5` — no null guard on `document.getElementById('discourse-comments')`; a page missing the div throws before the iframe is created.
- Test coverage gaps: no tests for `import_remote`/Readability, the real `poll_feed` body, XSS/raw_html sanitization, `absolutize_urls` port/protocol-relative cases, or `perform_retrieve` branches.

#### 💡 Suggestions
- `best.html.erb:6` — `<%- end if %>` is a confusing leftover (Ruby parses it as a modifier `if` whose condition is the next emitted buffer-append, so it renders identically to plain `end`); simplify to `<%- end %>`.

### Verdict
Recommend changes before merge — multiple stored-XSS sinks and a command-injection vector are security-blockers; the feed-poll nil-crash, transaction-boundary violation, and infinite loading loop are correctness-blockers.

app/models/post.rb
```diff
@@ -124,6 +128,11 @@ def post_analyzer
   end
 
   def cook(*args)
+    # For some posts, for example those imported via RSS, we support raw HTML. In that
+    # case we can skip the rendering pipeline.
+    return raw if cook_method == Post.cook_methods[:raw_html]
```
ron-x5labs, line 133: **Stored XSS via raw_html cook bypass.** `cook_method == raw_html` returns `raw` unchanged, skipping the `PrettyText.cook` sanitization/whitelist pipeline that normal posts pass through. The resulting `cooked` column is rendered raw in the embed iframe (`best.html.erb:19 <%= raw post.cooked %>`) **and on the main Discourse forum** (`post.js.handlebars:70 {{{cooked}}}`). Imported RSS/Readability HTML flows straight into this column with `skip_validations: true` → stored XSS that reaches the main forum, not just the embed. Since raw_html is meant to be an internal-only fast path, the imported content must still be sanitized before persistence (e.g. run it through `PrettyText`/`Sanitize` or a whitelist).

```ruby
def cook(*args)
  if cook_method == Post.cook_methods[:raw_html]
    # still sanitize imported HTML
    return Sanitize.clean(raw, Sanitize::Config::RELAXED)
  end
  Plugin::Filter.apply(:after_post_cook, self, post_analyzer.cook(*args))
end
```

app/models/topic_embed.rb
ron-x5labs, line 13: **Stored XSS via attribute breakout.** `url` is interpolated unescaped into `<a href='#{url}'>#{url}</a>`. The `/^https?:///` guard on line 11 has no end anchor and no `URI.parse`, so a valid-URI payload like `http://evil.com/'onclick='alert(1)` passes the regex (`'`/`(`/`)` are RFC3986 sub-delims) and breaks out of the single-quoted `href` attribute. The built string is then stored as `raw_html` cooked content (`skip_validations: true`) and rendered raw. Escape the URL and validate it fully:

```ruby
return unless url =~ /^https?:///\z/
uri = URI.parse(url)
raise Discourse::InvalidParameters if uri.host.blank?
safe_url = ERB::Util.html_escape(url)
contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: \"<a href='#{safe_url}'>#{safe_url}</a>\")}</small>\n"
```

app/models/topic_embed.rb
ron-x5labs, line 48: **Command injection / local file read via `Kernel#open`.** `import_remote` calls `open(url).read` *before* `import`'s `^https?://` scheme guard runs, and `TopicRetriever#invalid_host?` only compares `URI(url).host` (no scheme check). `Kernel#open` executes strings beginning with `|` as shell commands and honors `file://` URIs. The disqus thor task passes `t[:link]` from an XML export straight into `open(url)`, so a malicious/malformed export yields RCE / local file read. open-uri also follows redirects with no limit and no timeout (SSRF to internal hosts). Validate scheme+host before fetching and use a bounded fetch:

```ruby
def self.import_remote(user, url, opts=nil)
  raise Discourse::InvalidParameters if url !~ /^https?:///\z/
  uri = URI.parse(url)
  raise Discourse::InvalidParameters if uri.host != SiteSetting.embeddable_host
  require 'ruby-readability'
  doc = Readability::Document.new(open(url, read_timeout: 10, open_timeout: 10, redirect: false).read,
    tags: %w[div p code pre h1 h2 h3 b em i strong a img], attributes: %w[href src])
  TopicEmbed.import(user, url, opts[:title] || doc.title, doc.content)
rescue OpenURI::HTTPError, SocketError, Timeout::Error => e
  Rails.logger.warn("TopicEmbed#import_remote failed for #{url}: #{e.message}")
  nil
end
```

app/jobs/scheduled/poll_feed.rb
ron-x5labs, line 35: **`i.content.scrub` crashes for items lacking a `content` element.** SimpleRSS returns nil/raises for tags absent from a feed item; many RSS 2.0 feeds only carry `<description>`. `nil.scrub` raises `NoMethodError`, and because the `each` loop has no per-item rescue, one such item aborts the whole batch. With `sidekiq_options retry: false` that hour's feed import is silently lost. Guard nil and isolate each item:

```ruby
rss.items.each do |i|
  begin
    url = i.link
    url = i.id if url.blank? || url !~ /^https?:///\z/
    content = (i.content || i.description || '').to_s.scrub
    TopicEmbed.import(user, url, i.title, CGI.unescapeHTML(content))
  rescue => e
    Rails.logger.warn("PollFeed: skipping item #{i.try(:link)}: #{e.message}")
  end
end
```

app/views/embed/loading.html.erb
```diff
@@ -0,0 +1,12 @@
...
+      document.location.reload();
```
ron-x5labs, line 9: **Infinite reload loop.** The unconditional 30s `setTimeout(reload)` has no error state or retry cap. If the enqueued `RetrieveTopic` job failed, was throttled (60s `retrieved_recently?` guard), or the user is blank, the reload re-enters `embed#best`, finds no `topic_id`, re-enqueues another job, and re-renders `loading` — an infinite loop that always shows "Loading Discussion..." and hammers the server. Track attempt count and show a terminal failure state after N tries:

```js
(function() {
  var attempts = +(sessionStorage.getItem('discourse-embed-retries') || 0);
  if (attempts >= 5) { return; } // stop, leave the loading message
  sessionStorage.setItem('discourse-embed-retries', attempts + 1);
  setTimeout(function() { document.location.reload(); }, 30000);
})();
```

db/migrate/20131223171005_create_top_topics.rb
ron-x5labs, line 3: **`force: true` added to a previously-shipped migration.** `CreateTopTopics` already ran in existing environments; editing it to add `force: true` means a re-run (`db:migrate:redo` or a flaky deploy) drops and recreates `top_topics`, destroying all top-topic data. This change is also unrelated to the embed feature. Revert this to plain `create_table :top_topics do |t|` — `force` is unnecessary for a brand-new table on a fresh install and destructive on an existing one.

app/models/topic_embed.rb
ron-x5labs, line 21: **TOCTOU race + enqueue-before-commit.** `embed` is read outside any transaction (line 15), so concurrent retrieve/poll callers both enter the create branch and the loser's `TopicEmbed.create!` raises unhandled `RecordNotUnique` on the unique `embed_url` index. Wrapping `PostCreator.create` (which enqueues `:process_post`/`:feature_topic_users` *after its own transaction*, `post_creator.rb:288-294`) in an outer `Topic.transaction` breaks that guarantee: those jobs fire before the outer commit, and if `TopicEmbed.create!` fails they run against rolled-back ids → orphan jobs. Move the enqueue-after-commit responsibility out of the outer transaction, or rescue `RecordNotUnique` and re-fetch the winning embed:

```ruby
embed = TopicEmbed.where(embed_url: url).first
content_sha1 = Digest::SHA1.hexdigest(contents)
if embed.blank?
  post = PostCreator.new(user, title: title, raw: absolutize_urls(url, contents),
                         skip_validations: true, cook_method: Post.cook_methods[:raw_html]).create
  if post&.id.present? && post.errors.empty?
    begin
      TopicEmbed.create!(topic_id: post.topic_id, embed_url: url, content_sha1: content_sha1, post_id: post.id)
    rescue ActiveRecord::RecordNotUnique
      post = TopicEmbed.where(embed_url: url).first.post
    end
  end
else
  ...
end
```

lib/tasks/disqus.thor
ron-x5labs, line 148: **Breaking CLI change.** The documented `--category`/`-c` option was removed, and switching from `PostCreator` to `TopicEmbed.import_remote` no longer honors the thread's original `created_at` for the first post (replies keep theirs), producing topics dated after their own replies. Callers passing `--category` now fail with an unknown-option error. If dropping `--category` is intentional, thread `created_at` through `import_remote` -> `import` -> `PostCreator` (which already supports `created_at`):

```ruby
post = TopicEmbed.import_remote(user, t[:link], title: t[:title], created_at: Date.parse(t[:created_at]))
```

lib/topic_retriever.rb
ron-x5labs, line 27: **Non-atomic Redis throttle.** `$redis.setnx` + `$redis.expire` are two round-trips; a crash between them leaves the key with no TTL, permanently blocking that `embed_url`. Use a single atomic command:

```ruby
def retrieved_recently?
  return false if @opts[:no_throttle]
  !$redis.set("retrieved:#{@embed_url}", "1", ex: 60, nx: true)
end
```

lib/topic_retriever.rb
ron-x5labs, line 41: **Synchronous full feed poll inside the retrieve worker.** `Jobs::PollFeed.new.execute({})` runs the entire feed poll synchronously inside the per-request retrieve worker — blocking it, duplicating the hourly scheduled PollFeed, and amplifying the TOCTOU race on `TopicEmbed.import`. If RSS polling is the intended fast path, check the existing feed state without re-running the whole poll, or enqueue `PollFeed` asynchronously.

app/assets/javascripts/embed.js
ron-x5labs, line 17: **Substring origin check.** `discourseUrl.indexOf(e.origin) === -1` accepts any message whose origin is a *substring* of `discourseUrl` rather than an exact match — a known postMessage anti-pattern (CWE-349). Compare against the exact expected origin:

```js
var expectedOrigin = discourseUrl.match(/^https?:\/\/[^\/]+/)[0];
if (e.origin !== expectedOrigin) { return; }
```

app/views/layouts/embed.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<html>
+  <head>
+    <%= stylesheet_link_tag 'embed' %>
+
+    <script>
+      (function() {
+        window.onload = function() {
+          if (parent) {
+            // Send a post message with our loaded height
+            parent.postMessage({type: 'discourse-resize', height: document['body'].offsetHeight}, '<%= request.referer %>');
```
ron-x5labs, line 11: **postMessage targetOrigin is the raw Referer.** `request.referer` is a full URL with a path, but `postMessage` targetOrigin must be an origin (scheme+host+port), so the resize message is silently dropped in the common case (the iframe never resizes). Referer is also attacker-influenced. Use the trusted, validated origin (the controller already checks the referer host against `embeddable_host`):

```ruby
# in embed_controller#best, before render:
@embed_origin = (request.referer && URI(request.referer).scheme) ?
  URI(request.referer).scheme + "://" + URI(request.referer).host : SiteSetting.embeddable_host
```
```erb
parent.postMessage({type: 'discourse-resize', height: document.body.offsetHeight}, '<%= @embed_origin %>');
```

app/controllers/embed_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    discourse_expires_in 1.minute
+  end
+
+  private
+
+    def ensure_embeddable
+      raise Discourse::InvalidAccess.new('embeddable host not set') if SiteSetting.embeddable_host.blank?
+      raise Discourse::InvalidAccess.new('invalid referer host') if URI(request.referer || '').host != SiteSetting.embeddable_host
```
ron-x5labs, line 26: **Referer-only authorization.** The only gate is `URI(request.referer).host != embeddable_host` — the Referer header is client-supplied and forgeable/strippable, and `X-Frame-Options: ALLOWALL` lets any origin frame the endpoint. The `embed_url` param itself is never host-checked in the controller (only the enqueued job checks the host), so a stale `topic_id` for any embed_url renders a TopicView regardless of referer. Consider also validating `embed_url`'s host here and logging denied attempts (currently a silent 403 with no audit trail).

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+                                        attributes: %w[href src])
+
+    TopicEmbed.import(user, url, opts[:title] || doc.title, doc.content)
+  end
+
+  # Convert any relative URLs to absolute. RSS is annoying for this.
+  def self.absolutize_urls(url, contents)
+    uri = URI(url)
+    prefix = "#{uri.scheme}://#{uri.host}"
+    prefix << ":#{uri.port}" if uri.port != 80 && uri.port != 443
```
ron-x5labs, line 59: **Scheme-blind port logic.** `uri.port != 80 && uri.port != 443` ignores the scheme, so `https://host:80` and `http://host:443` lose their explicit port and resolve to the scheme default (a different endpoint). Compare against the scheme's default port:

```ruby
prefix << ":#{uri.port}" unless uri.port == uri.default_port
```

app/models/topic_embed.rb
ron-x5labs, line 64: **absolutize_urls mishandles protocol-relative and bare-relative URLs.** Only `start_with?('/')` hrefs/srcs are rewritten, and `sub(/^\/+/, '')` strips all leading slashes, so `//cdn.host/x` becomes `http://<embed-host>/cdn.host/x` (a broken embed-host path). Bare-relative links (`foo/bar`) are left untouched and resolve against Discourse, not the source. Use `URI.join` against the base URL, which handles `//`, `/`, and bare-relative cases:

```ruby
fragment.css('a').each do |a|
  href = a['href']
  next if href.blank? || href =~ /^https?:\/\//i
  a['href'] = URI.join(url, href).to_s
end
```

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+    embed = TopicEmbed.where(embed_url: url).first
+    content_sha1 = Digest::SHA1.hexdigest(contents)
+    post = nil
+
+    # If there is no embed, create a topic, post and the embed.
+    if embed.blank?
+      Topic.transaction do
+        creator = PostCreator.new(user, title: title, raw: absolutize_urls(url, contents), skip_validations: true, cook_method: Post.cook_methods[:raw_html])
+        post = creator.create
+        if post.present?
```
ron-x5labs, line 24: **`post.present?` is true after a rolled-back create.** `PostCreator.create` returns the in-memory `@post` even when the save rolled back (`errors` set, `id`/`topic_id` nil) — e.g. on `rollback_if_host_spam_detected`. `post.present?` is still true, so `TopicEmbed.create!` runs with nil `topic_id`/`post_id` and hits the NOT NULL constraints. Check persistence instead:

```ruby
if post&.id.present? && post.errors.empty?
  TopicEmbed.create!(topic_id: post.topic_id, embed_url: url, content_sha1: content_sha1, post_id: post.id)
end
```

app/jobs/scheduled/poll_feed.rb
ron-x5labs, line 29: **No error handling + `retry: false` silently drops the hourly poll.** `open(feed_polling_url)` and `SimpleRSS.parse` have no `rescue`; a transient network/DNS/parse error aborts the job, and with `retry: false` it never self-heals — no application-level log, no alert. Wrap the fetch/parse in a `rescue` that logs with the feed URL:

```ruby
begin
  require 'simple-rss'
  rss = SimpleRSS.parse(open(SiteSetting.feed_polling_url, read_timeout: 10, open_timeout: 10))
rescue => e
  Rails.logger.warn("PollFeed failed for #{SiteSetting.feed_polling_url}: #{e.class} #{e.message}")
  raise
end
```

app/controllers/embed_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  before_filter :ensure_embeddable
+
+  layout 'embed'
+
+  def best
+    embed_url = params.require(:embed_url)
+    topic_id = TopicEmbed.topic_id_for_embed(embed_url)
+
+    if topic_id
+      @topic_view = TopicView.new(topic_id, current_user, {best: 5})
```
ron-x5labs, line 13: **`TopicView.new` raises `NotFound` for a stale embed.** When `topic_id_for_embed` returns an id whose topic was deleted or is inaccessible, `TopicView#check_and_raise_exceptions` raises `Discourse::NotFound` (`topic_view.rb:300`), so the embed iframe shows an error page instead of falling back to the loading/retrieve path. Wrap and fall back:

```ruby
begin
  @topic_view = TopicView.new(topic_id, current_user, {best: 5})
rescue Discourse::NotFound
  @topic_view = nil
end
if @topic_view.nil?
  Jobs.enqueue(:retrieve_topic, user_id: current_user.try(:id), embed_url: embed_url)
  render 'loading'
end
```

app/jobs/scheduled/poll_feed.rb
```diff
@@ -0,0 +1,41 @@
...
+    recurrence { hourly }
+    sidekiq_options retry: false
+
+    def execute(args)
+      poll_feed if SiteSetting.feed_polling_enabled? &&
+                   SiteSetting.feed_polling_url.present? &&
+                   SiteSetting.embed_by_username.present?
+    end
+
+    def feed_key
```
ron-x5labs, line 20: **`feed_key` is dead code.** It is defined but never referenced anywhere in the codebase, implying an intended modified-since dedup that was never wired. As a result the full feed is re-fetched and re-iterated every hour. Either wire it (e.g. cache the last-modified/etag and skip unchanged feeds) or remove it to avoid confusion.

Gemfile
```diff
@@ -206,6 +206,10 @@ gem 'unicorn', require: false
 gem 'puma', require: false
 gem 'rbtrace', require: false
 
+# required for feed importing and embedding
+gem 'ruby-readability', require: false
```
ron-x5labs, line 210: **Gems added without version constraints and unmaintained.** `ruby-readability` and `simple-rss` have no version pin (only `Gemfile_rails4.lock` pins 0.5.7/1.3.1); the repo pins most other gems (`sidekiq`, '2.15.1', `pg`, '0.15.1'), so these break convention. Both gems are long-unmaintained (simple-rss ~2009), widening the unvetted dependency surface (ruby-readability also pulls in `guess_html_encoding`). Pin them: `gem 'ruby-readability', '~> 0.5.7'` / `gem 'simple-rss', '~> 1.3.1'`.

ron-x5labs (changes requested):
## Code Review: FEATURE: Embeddable Discourse comments, now with simple-rss instead of feedzirra

### Problem
Adds embeddable Discourse comments: a third-party site includes `embed.js`, which iframes `embed/best?embed_url=...`. The controller (referer-gated) renders a `TopicView` of the best 5 posts, or enqueues a `RetrieveTopic` job that fetches the source page via `ruby-readability` and stores it as a `cook_method=raw_html` post whose `cooked == raw` (bypasses the cook/sanitization pipeline). A scheduled `PollFeed` job imports an RSS/Atom feed the same way.

### Solution Reviewed
New `EmbedController` (referer-gated, `X-Frame-Options: ALLOWALL`), `TopicEmbed` model (import/create/revise + `absolutize_urls`), `TopicRetriever` (Redis-throttled), `RetrieveTopic`/`PollFeed` jobs, `Post#cook_methods` enum (`raw_html` short-circuits cook), migrations (`topic_embeds`, `posts.cook_method`), embed views/layout/JS/CSS, site settings, locales. Disqus thor rewired through `import_remote`.

### Summary
The feature is well-structured and the core wiring (enum default=1=regular, PostCreator opt, PostRevisor `skip_validations`, routes, settings, locales) is internally consistent. However it ships with **two stored-XSS chains** that reach the main forum (not just the embed iframe), a **command-injection/local-file-read** vector via `Kernel#open`, a **nil-crash that breaks feed import for common feeds**, a **transaction-boundary violation** that can orphan enqueued jobs, an **infinite reload loop**, and a **data-loss migration edit**. Recommend changes before merge.

### Files Reviewed
- `app/models/post.rb` — deeply reviewed (cook override)
- `app/models/topic_embed.rb` — deeply reviewed (import/import_remote/absolutize_urls)
- `app/controllers/embed_controller.rb` — deeply reviewed
- `app/jobs/scheduled/poll_feed.rb` — deeply reviewed
- `app/jobs/regular/retrieve_topic.rb` — deeply reviewed
- `lib/topic_retriever.rb` — deeply reviewed
- `lib/post_creator.rb` / `lib/post_revisor.rb` — deeply reviewed (touched)
- `lib/tasks/disqus.thor` — deeply reviewed
- `app/assets/javascripts/embed.js` — deeply reviewed
- `app/views/embed/best.html.erb` / `loading.html.erb` / `layouts/embed.html.erb` — deeply reviewed
- `db/migrate/*` (4 files) — deeply reviewed
- `config/routes.rb`, `config/site_settings.yml`, `config/locales/*.en.yml` — deeply reviewed
- `Gemfile`, `Gemfile_rails4.lock` — deeply reviewed (note: `Gemfile.lock`/`Gemfile_rails_master.lock` not updated)
- spec files (4) — lightly reviewed (coverage gaps noted)
- `app/assets/stylesheets/embed.css.scss` — lightly reviewed (styles only)

### Verification
- No Ruby/JS toolchain available in the review environment — typecheck/lint/tests/build skipped. Findings grounded in static analysis of the checked-out PR source (confirmed: `Enum.new(:regular, :raw_html)` => regular=1, raw_html=2, migration default=1; `post.js.handlebars:70` renders `{{{cooked}}}` raw on the main forum; `topic_view.rb:300` raises `Discourse::NotFound` for missing topic; `post_creator.rb:288-294` enqueues jobs after its inner transaction; `anonymous_cache.rb:14` cache key excludes Referer).

### Issues Found
See inline comments. Highlights:

#### Blocking
- **post.rb:133** — `cook_method=raw_html` returns `raw` unsanitized; `cooked` is rendered raw in the embed (`best.html.erb:19`) AND on the main forum (`post.js.handlebars:70 {{{cooked}}}`). Imported RSS/Readability HTML bypasses PrettyText sanitization entirely → stored XSS.
- **topic_embed.rb:13** — `url` interpolated unescaped into `<a href='#{url}'>`; the `^https?://` regex has no end anchor / no `URI.parse`, so `'`/`<`/`>` in the URL break out of the single-quoted attribute → stored XSS via the raw_html cooked column.
- **topic_embed.rb:48** — `import_remote` calls `open(url).read` *before* `import`'s scheme guard; `Kernel#open` executes `|`-prefixed strings as shell commands and reads `file://`. The disqus thor path passes `t[:link]` from an XML export straight to `open(url)` → command injection / local file read.
- **poll_feed.rb:35** — `i.content.scrub` raises `NoMethodError` for feed items lacking a `content` element (common — many feeds only carry `description`); aborts the whole `rss.items.each` loop. With `retry: false` the hourly poll is silently lost.
- **loading.html.erb:9** — unconditional 30s `setTimeout(reload)` with no error/retry cap; if the retrieve job failed or was throttled (60s guard), reload re-enters `embed#best`, finds no topic_id, re-enqueues, re-renders loading → infinite loop.
- **create_top_topics.rb:3** — a previously-shipped migration is edited to add `force: true`; on re-run (`db:migrate:redo` or a flaky deploy) this drops and recreates `top_topics`, destroying data. Unrelated to this feature.
- **topic_embed.rb:21** — `embed` read outside the transaction (line 15); concurrent retrieve/poll callers both enter the create branch and the loser's `TopicEmbed.create!` raises unhandled `RecordNotUnique`. Wrapping `PostCreator.create` (which enqueues `:process_post`/`:feature_topic_users` after its *own* transaction, `post_creator.rb:288-294`) in an outer `Topic.transaction` means those jobs fire before the outer commit; if `TopicEmbed.create!` fails they run against rolled-back ids → orphan jobs.
- **disqus.thor:148** — breaking CLI change: the documented `--category`/`-c` option is removed, and the thread's original `created_at` is no longer honored for the imported topic's first post (replies keep theirs), producing topics dated after their own replies.

#### Non-blocking
- **topic_retriever.rb:27** — `$redis.setnx` + `$redis.expire` are two round-trips; a crash between them leaves the key with no TTL, permanently blocking that embed_url. Use atomic `SET key 1 EX 60 NX`.
- **topic_retriever.rb:41** — `Jobs::PollFeed.new.execute({})` runs the full feed poll synchronously inside the retrieve worker (blocking, duplicates the scheduled PollFeed, amplifies the TOCTOU race).
- **embed.js:17** — `discourseUrl.indexOf(e.origin) === -1` is substring containment, not exact origin equality; compare `e.origin === new URL(discourseUrl).origin`.
- **layouts/embed.html.erb:11** — `postMessage` `targetOrigin` is `request.referer` (attacker-influenced, and a full URL with path ≠ origin, so the resize message is silently dropped in the common case). Should be a fixed expected origin.
- **embed_controller.rb:26** — authorization rests solely on the forgeable `Referer` header; `embed_url` itself is never host-checked in the controller, and denied/abusive embed attempts are not logged.
- **topic_embed.rb:59** — `uri.port != 80 && uri.port != 443` is scheme-blind; `https://host:80` and `http://host:443` lose their explicit port and resolve to the wrong endpoint. Use `uri.port == uri.default_port`.
- **topic_embed.rb:64** — `absolutize_urls` only rewrites `start_with?('/')` hrefs/srcs; protocol-relative `//host/x` is mangled to an embed-host path, and bare-relative links are left relative (resolve against Discourse, not source). Use `URI.join(url, href)`.
- **topic_embed.rb:24** — `post.present?` is true even when `PostCreator.create` rolled back (errors set, nil id/topic_id); `TopicEmbed.create!` then runs with nil ids and violates NOT NULL.
- **poll_feed.rb:29** — `open(SiteSetting.feed_polling_url)` + `SimpleRSS.parse` have no rescue and `retry: false`; a transient network/parse error silently drops the entire hourly poll with no logging.
- **embed_controller.rb:13** — `TopicView.new(topic_id, ...)` raises `Discourse::NotFound` when the `TopicEmbed` row is stale (topic deleted/inaccessible); the embed endpoint errors out instead of falling back to loading/retrieve.
- Stale lockfiles: `Gemfile.lock` and `Gemfile_rails_master.lock` were not regenerated (only `Gemfile_rails4.lock`); `bundle install --deployment` will fail or silently regenerate on the default lockfile.

#### Suggestions
- **poll_feed.rb:20** — `feed_key` is dead code (never referenced); the intended modified-since throttle is never wired, so the full feed is re-fetched/re-iterated hourly.
- **Gemfile:210** — `ruby-readability` and `simple-rss` are added without version constraints and both are long-unmaintained; the repo pins most other gems, so these break convention. Pin with `~> 0.5.7` / `~> 1.3.1`.
- Test coverage gaps: no tests for `import_remote`/Readability, the real `poll_feed` body, XSS/raw_html sanitization, `absolutize_urls` port/protocol-relative cases, or `perform_retrieve`/`retrieved_recently?` bodies (specs stub them out).

### Verdict
Recommend changes before merge — the stored-XSS chains and command-injection vector are security-blockers; the feed-poll nil-crash, transaction-boundary violation, infinite loading loop, and data-loss migration edit are correctness-blockers.

app/views/embed/best.html.erb
ron-x5labs, line 19: 🔴 **Stored XSS.** `<%= raw post.cooked %>` renders imported HTML unescaped. `cook_method=raw_html` makes `Post#cook` return `raw` unchanged (post.rb:133), bypassing the `PrettyText.cook` sanitize pipeline, so `cooked` is the raw RSS/scraped HTML — including any `<script>`/`onerror=` from a malicious or compromised feed. The iframe is served from the Discourse origin, so this is same-origin script execution against the viewer's session (and `cooked` is also rendered raw in the normal topic/email views). Sanitize imported content before storing (route it through `PrettyText.cook`/`Sanitize.clean`, or a strict Nokogiri allowlist) and don't short-circuit `Post#cook` to `raw` for raw_html posts.

app/models/topic_embed.rb
ron-x5labs, line 13: 🔴 **Stored XSS via attribute breakout.** `url` is interpolated unescaped into `<a href='#{url}'>#{url}</a>`. The guard at line 11 (`url =~ /^https?://`) only anchors the *start*, so a URL like `http://x' onclick='alert(1)//` closes the single-quoted attribute and injects a handler; since the post is `raw_html` the markup is stored verbatim and rendered via `raw`. `url` is attacker-influenced (RSS `i.link`/`i.id`, or `params[:embed_url]`). Validate with `URI.parse` + scheme check and escape on output:
```ruby
return unless (u = URI.parse(url) rescue nil) && %w[http https].include?(u.scheme)
# in the appended markup:
esc = ERB::Util.method(:html_escape)
contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: "<a href='#{esc.call(url)}'>#{esc.call(url)}</a>")}</small>\n"
```

app/models/topic_embed.rb
ron-x5labs, line 48: 🔴 **Command injection / local-file read via `Kernel#open`.** `open(url).read` runs *before* `import`'s `^https?://` scheme check (that check is inside `import`, line 11, not here). `Kernel#open` treats `|cmd` as a shell command and `file:` as local-file access; `import_remote` is reached from `disqus.thor` with `t[:link]` that is **not** host-checked, and open-uri follows HTTP redirects without re-validating the host (SSRF to internal services such as cloud metadata). Validate the scheme first and use `URI#read`:
```ruby
u = URI.parse(url)
raise Discourse::InvalidAccess unless %w[http https].include?(u.scheme)
doc = Readability::Document.new(u.read, tags: ..., attributes: ...)
```
Apply the same to `poll_feed.rb:29` (`open(SiteSetting.feed_polling_url)`).

app/models/topic_embed.rb
```diff
@@ -0,0 +1,82 @@
...
+  validates_presence_of :embed_url
+  validates_presence_of :content_sha1
+
+  # Import an article from a source (RSS/Atom/Other)
+  def self.import(user, url, title, contents)
+    return unless url =~ /^https?\:\/\//
+
+    contents << "\n<hr>\n<small>#{I18n.t('embed.imported_from', link: "<a href='#{url}'>#{url}</a>")}</small>\n"
+
+    embed = TopicEmbed.where(embed_url: url).first
```
ron-x5labs, line 15: 🔴 **TOCTOU race.** `embed` is read *outside* the transaction (line 15); `TopicEmbed.create!` happens inside it (line 25) against the unique index on `embed_url`. Concurrent callers — `EmbedController#best` enqueues `RetrieveTopic` while the hourly `PollFeed` iterates the same URLs on separate workers — both observe `embed.blank?`, both create, and the loser raises `RecordNotUnique`, which propagates unhandled and aborts the rest of the `poll_feed` item loop. Move the lookup inside the transaction, or `rescue ActiveRecord::RecordNotUnique` and re-read the embed to take the update branch.

app/jobs/scheduled/poll_feed.rb
ron-x5labs, line 35: 🔴 **PollFeed is broken on the documented runtime and on common feeds.** `i.content.scrub` raises `NoMethodError` two ways: `String#scrub` was added in Ruby 2.1.0 but `.travis.yml` pins Ruby 2.0.0 (no `string-scrub` backport in any lockfile), and `i.content` is `nil` for any RSS item lacking `<content:encoded>` (most RSS 2.0 feeds expose only `<description>`). With `sidekiq_options retry: false` (line 12) the entire hourly import silently fails. Nil-guard and fall back to description:
```ruby
content = i.content || i.description || ''
content = CGI.unescapeHTML(content.to_s.scrub)
```
(and require/drop `scrub` to support Ruby 2.0.0).

app/views/embed/loading.html.erb
ron-x5labs, line 9: 🔴 **The loading→best auto-transition is broken.** `document.location.reload()` resends the iframe's own Discourse URL as `Referer`, so on the next request `ensure_embeddable` sees `URI(request.referer).host != embeddable_host` and raises `InvalidAccess` (a 403 error page, not the embed layout). The 1-minute anonymous cache only delays this — once it expires the embed 403s instead of showing the discussion. There's also no retry cap or error state, so a failed/throttled `RetrieveTopic` job reloads forever. Drive the reload from the parent `embed.js` (which knows the real embed page Referer), and render a terminal error state after N attempts.

db/migrate/20131223171005_create_top_topics.rb
ron-x5labs, line 3: 🔴 **`force: true` on a previously-shipped migration.** `CreateTopTopics` already ran in existing environments; editing it to add `force: true` means `db:migrate:redo` (or a flaky re-applying deploy) issues `DROP TABLE top_topics` and destroys all computed top-topic aggregates for every period. `force: true` is a dev-iteration aid and must not ship. Revert to `create_table :top_topics do |t|` (and note `create_topic_embeds` ships `force: true` too — harmless for a brand-new table but should also be dropped for consistency).

app/models/topic_embed.rb
ron-x5labs, line 24: 🔴 **`post.present?` is true after a rolled-back create.** `PostCreator#create` returns the in-memory `@post` even when the save rolled back (`errors` set, `id`/`topic_id` nil — verified: `create` returns `@post` after the `Post.transaction do ... end` block). So `post.present?` passes and `TopicEmbed.create!(topic_id: nil, post_id: nil, ...)` raises a DB NOT-NULL violation (the model only validates `embed_url`/`content_sha1`, not `topic_id`/`post_id`). Test persistence, not presence:
```ruby
if post.present? && post.id.present?
```

lib/topic_retriever.rb
ron-x5labs, line 27: 🟡 **Non-atomic Redis throttle.** `$redis.setnx` + `$redis.expire` are two round-trips; a crash/kill between them leaves `retrieved:<embed_url>` with no TTL, permanently blocking retrieval for that URL until an operator `DEL`s it. Use a single atomic command:
```ruby
return $redis.set(retrieved_key, '1', nx: true, ex: 60) ? false : true
```

app/jobs/scheduled/poll_feed.rb
ron-x5labs, line 29: 🟡 **No error handling + `retry: false` silently drops the poll.** `open(feed_polling_url)` and `SimpleRSS.parse` raise on transient network/DNS/SSL/parse errors with no `rescue`, and a single failing item in the loop (line 31) aborts every subsequent item; with `retry: false` (line 12) the hourly poll is dropped and, since the failing item keeps its position, everything after it is permanently skipped each run. Wrap the fetch in a rescue that logs feed-level failures, and isolate each item (`begin/rescue => e; Rails.logger.warn(...); next; end`).

lib/topic_retriever.rb
ron-x5labs, line 41: 🟡 **Synchronous full feed poll inside the retrieve worker.** `Jobs::PollFeed.new.execute({})` runs the entire feed fetch+parse+import inline inside the per-request `RetrieveTopic` worker, blocking it and duplicating work. It also bypasses the scheduled-job lock, so a Sidetiq hourly tick racing this inline call double-processes items (compounding the `import` TOCTOU). Enqueue it instead: `Jobs.enqueue(:poll_feed)`.

lib/tasks/disqus.thor
ron-x5labs, line 148: 🟡 **Breaking Disqus CLI change.** `--category`/`-c` was removed, so existing invocations passing it now error with an unknown-option Thor failure; and switching to `TopicEmbed.import_remote` no longer forwards `created_at`, so imported topics get `Time.now` while their replies still use `Date.parse(p[:created_at])` (~line 164) — replies can predate their own topic. Thread `created_at`/`category` through `TopicEmbed.import` → `PostCreator` (which already honors both).

app/controllers/embed_controller.rb
ron-x5labs, line 13: 🟡 **Stale/orphaned embed → permanent 404.** When an imported topic is deleted, `topic_id_for_embed` still returns its id (no `dependent: :destroy` on the Topic side), so `TopicView.new` raises `Discourse::NotFound`/`InvalidAccess`, rendered as a raw error page inside the iframe; the `else` re-import branch is only taken when `topic_id` is falsy. Rescue/nil-check the `TopicView` and fall through to retrieval, or add `dependent: :destroy` so deleting the topic removes the `topic_embeds` row.

app/models/topic_embed.rb
ron-x5labs, line 64: 🟡 **`absolutize_urls` mangles protocol-relative URLs.** `//cdn.example.com/x` satisfies `start_with?('/')` and `sub(/^\/+/, '')` strips *both* leading slashes, yielding `http://<embedhost>/cdn.example.com/x` (wrong host). Exclude protocol-relative URLs (and note `:uri.port != 80 && != 443` at line 59 ignores scheme — `http://host:443`/`https://host:80` lose their explicit port; prefer `uri.port != uri.default_port`):
```ruby
if href.present? && href.start_with?('/') && !href.start_with?('//')
```

Gemfile
ron-x5labs, line 210: 🟡 **Unpinned gems + only one lockfile regenerated.** `ruby-readability`/`simple-rss` have no version constraint (the repo pins most gems), so a future resolve can pull an unmaintained/typosquat release. And the diff only updates `Gemfile_rails4.lock` — `Gemfile.lock` and `Gemfile_rails_master.lock` are missing these gems (and transitive `guess_html_encoding`), so rails3/master `bundle check` is out of sync. Pin versions (`gem 'ruby-readability', '0.5.7'`) and regenerate all three lockfiles.

app/jobs/scheduled/poll_feed.rb
ron-x5labs, line 20: 💡 **`feed_key` is dead code.** Defined but never referenced anywhere — it looks like an intended Last-Modified/ETag conditional-GET throttle that was never wired up, so every hourly tick re-fetches and re-imports the whole feed with no idempotency guard. Either wire it up or remove it.

config/locales/server.en.yml
```diff
@@ -29,6 +29,14 @@ en:
 
   too_many_replies: "Sorry you can't reply any more times in that topic."
 
+  embed:
+    title: "Discussion Highlights"
+    start_discussion: "Begin the Discussion"
+    continue: "Continue the Discussion"
+    loading: "Loading Discussion..."
+    permalink: "Permalink"
```
ron-x5labs, line 37: 💡 `embed.permalink` is defined but no view/controller/model references it. Remove it or add the consuming UI.

PR #8, opened 2025-07-26
FIX: proper handling of group memberships

everettbu:
Test 8

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-group.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-group.js.es6
@@ -1,34 +1,69 @@
 export default Em.ObjectController.extend({
   needs: ['adminGroups'],
-  members: null,
   disableSave: false,
+  usernames: null,
+
+  currentPage: function() {
+    if (this.get("user_count") == 0) { return 0; }
+    return Math.floor(this.get("offset") / this.get("limit")) + 1;
+  }.property("limit", "offset", "user_count"),
+
+  totalPages: function() {
+    if (this.get("user_count") == 0) { return 0; }
+    return Math.floor(this.get("user_count") / this.get("limit")) + 1;
+  }.property("limit", "user_count"),
+
+  showingFirst: Em.computed.lte("currentPage", 1),
+  showingLast: Discourse.computed.propertyEqual("currentPage", "totalPages"),
 
   aliasLevelOptions: function() {
     return [
-      { name: I18n.t("groups.alias_levels.nobody"), value: 0},
-      { name: I18n.t("groups.alias_levels.mods_and_admins"), value: 2},
-      { name: I18n.t("groups.alias_levels.members_mods_and_admins"), value: 3},
-      { name: I18n.t("groups.alias_levels.everyone"), value: 99}
+      { name: I18n.t("groups.alias_levels.nobody"), value: 0 },
+      { name: I18n.t("groups.alias_levels.mods_and_admins"), value: 2 },
+      { name: I18n.t("groups.alias_levels.members_mods_and_admins"), value: 3 },
+      { name: I18n.t("groups.alias_levels.everyone"), value: 99 }
     ];
   }.property(),
 
-  usernames: function(key, value) {
-    var members = this.get('members');
-    if (arguments.length > 1) {
-      this.set('_usernames', value);
-    } else {
-      var usernames;
-      if(members) {
-        usernames = members.map(function(user) {
-          return user.get('username');
-        }).join(',');
-      }
-      this.set('_usernames', usernames);
-    }
-    return this.get('_usernames');
-  }.property('members.@each.username'),
-
   actions: {
+    next: function() {
+      if (this.get("showingLast")) { return; }
+
+      var group = this.get("model"),
+          offset = Math.min(group.get("offset") + group.get("limit"), group.get("user_count"));
+
+      group.set("offset", offset);
+
+      return group.findMembers();
+    },
+
+    previous: function() {
+      if (this.get("showingFirst")) { return; }
+
+      var group = this.get("model"),
+          offset = Math.max(group.get("offset") - group.get("limit"), 0);
+
+      group.set("offset", offset);
+
+      return group.findMembers();
+    },
+
+    removeMember: function(member) {
+      var self = this,
+          message = I18n.t("admin.groups.delete_member_confirm", { username: member.get("username"), group: this.get("name") });
+      return bootbox.confirm(message, I18n.t("no_value"), I18n.t("yes_value"), function(confirm) {
+        if (confirm) {
+          self.get("model").removeMember(member);
+        }
+      });
+    },
+
+    addMembers: function() {
+      // TODO: should clear the input
+      if (Em.isEmpty(this.get("usernames"))) { return; }
+      this.get("model").addMembers(this.get("usernames"));
+    },
+
     save: function() {
       var self = this,
           group = this.get('model');
@@ -37,9 +72,9 @@ export default Em.ObjectController.extend({
 
       var promise;
       if (group.get('id')) {
-        promise = group.saveWithUsernames(this.get('usernames'));
+        promise = group.save();
       } else {
-        promise = group.createWithUsernames(this.get('usernames')).then(function() {
+        promise = group.create().then(function() {
           var groupsController = self.get('controllers.adminGroups');
           groupsController.addObject(group);
         });
--- a/app/assets/javascripts/discourse/models/group.js
+++ b/app/assets/javascripts/discourse/models/group.js
@@ -7,53 +7,80 @@
   @module Discourse
 **/
 Discourse.Group = Discourse.Model.extend({
+  limit: 50,
+  offset: 0,
+  user_count: 0,
 
   userCountDisplay: function(){
     var c = this.get('user_count');
     // don't display zero its ugly
-    if(c > 0) {
-      return c;
-    }
+    if (c > 0) { return c; }
   }.property('user_count'),
 
   findMembers: function() {
-    if (Em.isEmpty(this.get('name'))) { return Ember.RSVP.resolve([]); }
+    if (Em.isEmpty(this.get('name'))) { return ; }
 
-    return Discourse.ajax('/groups/' + this.get('name') + '/members').then(function(result) {
-      return result.map(function(u) { return Discourse.User.create(u) });
+    var self = this, offset = Math.min(this.get("user_count"), Math.max(this.get("offset"), 0));
+
+    return Discourse.ajax('/groups/' + this.get('name') + '/members.json', {
+      data: {
+        limit: this.get("limit"),
+        offset: offset
+      }
+    }).then(function(result) {
+      self.setProperties({
+        user_count: result.meta.total,
+        limit: result.meta.limit,
+        offset: result.meta.offset,
+        members: result.members.map(function(member) { return Discourse.User.create(member); })
+      });
     });
   },
 
-  destroy: function(){
-    if(!this.get('id')) return;
-    return Discourse.ajax("/admin/groups/" + this.get('id'), {type: "DELETE"});
+  removeMember: function(member) {
+    var self = this;
+    return Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {
+      type: "DELETE",
+      data: { user_id: member.get("id") }
+    }).then(function() {
+      // reload member list
+      self.findMembers();
+    });
+  },
+
+  addMembers: function(usernames) {
+    var self = this;
+    return Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {
+      type: "PUT",
+      data: { usernames: usernames }
+    }).then(function() {
+      // reload member list
+      self.findMembers();
+    })
   },
 
   asJSON: function() {
-    return { group: {
-             name: this.get('name'),
-             alias_level: this.get('alias_level'),
-             visible: !!this.get('visible'),
-             usernames: this.get('usernames') } };
+    return {
+      name: this.get('name'),
+      alias_level: this.get('alias_level'),
+      visible: !!this.get('visible')
+    };
   },
 
-  createWithUsernames: function(usernames){
-    var self = this,
-        json = this.asJSON();
-    json.group.usernames = usernames;
-
-    return Discourse.ajax("/admin/groups", {type: "POST", data: json}).then(function(resp) {
+  create: function(){
+    var self = this;
+    return Discourse.ajax("/admin/groups", { type: "POST", data: this.asJSON() }).then(function(resp) {
       self.set('id', resp.basic_group.id);
     });
   },
 
-  saveWithUsernames: function(usernames){
-    var json = this.asJSON();
-    json.group.usernames = usernames;
-    return Discourse.ajax("/admin/groups/" + this.get('id'), {
-      type: "PUT",
-      data: json
-    });
+  save: function(){
+    return Discourse.ajax("/admin/groups/" + this.get('id'), { type: "PUT", data: this.asJSON() });
+  },
+
+  destroy: function(){
+    if (!this.get('id')) { return };
+    return Discourse.ajax("/admin/groups/" + this.get('id'), {type: "DELETE"});
   },
 
   findPosts: function(opts) {
--- a/app/controllers/admin/groups_controller.rb
+++ b/app/controllers/admin/groups_controller.rb
@@ -1,13 +1,17 @@
 class Admin::GroupsController < Admin::AdminController
+
   def index
     groups = Group.order(:name)
+
     if search = params[:search]
       search = search.to_s
-      groups = groups.where("name ilike ?", "%#{search}%")
+      groups = groups.where("name ILIKE ?", "%#{search}%")
     end
+
     if params[:ignore_automatic].to_s == "true"
       groups = groups.where(automatic: false)
     end
+
     render_serialized(groups, BasicGroupSerializer)
   end
 
@@ -15,45 +19,13 @@ class Admin::GroupsController < Admin::AdminController
     render nothing: true
   end
 
-  def refresh_automatic_groups
-    Group.refresh_automatic_groups!
-    render json: success_json
-  end
-
-  def update_patch(group)
-    raise Discourse::InvalidAccess.new("automatic groups do not permit membership changes") if group.automatic
-
-    if actions = params[:changes]
-      Array(actions[:add]).each do |username|
-        if user = User.find_by_username(username)
-          group.add(user)
-        end
-      end
-      Array(actions[:delete]).each do |username|
-        if user = User.find_by_username(username)
-          group.remove(user)
-        end
-      end
-    end
-
-    render json: success_json
-  end
-
-  def update_put(group)
-    payload = params[:group]
-
-    group.alias_level = payload[:alias_level].to_i if payload[:alias_level].present?
-    group.visible = payload[:visible] == "true"
-
-    if group.automatic
-      # group rename & membership changes are ignored/prohibited for automatic groups
-    else
-      group.usernames = payload[:usernames] if payload[:usernames]
-      group.name = payload[:name] if payload[:name]
-    end
+  def create
+    group = Group.new
+    group.name = (params[:name] || '').strip
+    group.visible = params[:visible] == "true"
 
     if group.save
-      render json: success_json
+      render_serialized(group, BasicGroupSerializer)
     else
       render_json_error group
     end
@@ -62,20 +34,13 @@ class Admin::GroupsController < Admin::AdminController
   def update
     group = Group.find(params[:id].to_i)
 
-    if request.patch?
-      update_patch(group)
-    else
-      update_put(group)
-    end
-  end
+    group.alias_level = params[:alias_level].to_i if params[:alias_level].present?
+    group.visible = params[:visible] == "true"
+    # group rename is ignored for automatic groups
+    group.name = params[:name] if params[:name] && !group.automatic
 
-  def create
-    group = Group.new
-    group.name = (params[:group][:name] || '').strip
-    group.usernames = params[:group][:usernames] if params[:group][:usernames]
-    group.visible = params[:group][:visible] == "true"
     if group.save
-      render_serialized(group, BasicGroupSerializer)
+      render json: success_json
     else
       render_json_error group
     end
@@ -83,6 +48,7 @@ class Admin::GroupsController < Admin::AdminController
 
   def destroy
     group = Group.find(params[:id].to_i)
+
     if group.automatic
       can_not_modify_automatic
     else
@@ -91,9 +57,48 @@ class Admin::GroupsController < Admin::AdminController
     end
   end
 
+  def refresh_automatic_groups
+    Group.refresh_automatic_groups!
+    render json: success_json
+  end
+
+  def add_members
+    group = Group.find(params.require(:group_id).to_i)
+    usernames = params.require(:usernames)
+
+    return can_not_modify_automatic if group.automatic
+
+    usernames.split(",").each do |username|
+      if user = User.find_by_username(username)
+        group.add(user)
+      end
+    end
+
+    if group.save
+      render json: success_json
+    else
+      render_json_error(group)
+    end
+  end
+
+  def remove_member
+    group = Group.find(params.require(:group_id).to_i)
+    user_id = params.require(:user_id).to_i
+
+    return can_not_modify_automatic if group.automatic
+
+    group.users.delete(user_id)
+
+    if group.save
+      render json: success_json
+    else
+      render_json_error(group)
+    end
+  end
+
   protected
 
-  def can_not_modify_automatic
-    render json: {errors: I18n.t('groups.errors.can_not_modify_automatic')}, status: 422
-  end
+    def can_not_modify_automatic
+      render json: {errors: I18n.t('groups.errors.can_not_modify_automatic')}, status: 422
+    end
 end
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -19,26 +19,29 @@ class GroupsController < ApplicationController
   def members
     group = find_group(:group_id)
 
-    members = group.users.order('username_lower asc')
+    limit = (params[:limit] || 50).to_i
+    offset = params[:offset].to_i
 
-    # TODO: We should fix the root cause of the bug where if there
-    # are more than 200 groups it will truncate
-    if group.automatic?
-      limit = (params[:limit] || 200).to_i
-      offset = (params[:offset] || 0).to_i
-      members = members.limit(limit).offset(offset)
-    end
+    total = group.users.count
+    members = group.users.order(:username_lower).limit(limit).offset(offset)
 
-    render_serialized(members.to_a, GroupUserSerializer)
+    render json: {
+      members: serialize_data(members, GroupUserSerializer),
+      meta: {
+        total: total,
+        limit: limit,
+        offset: offset
+      }
+    }
   end
 
   private
 
-  def find_group(param_name)
-    name = params.require(param_name)
-    group = Group.find_by("lower(name) = ?", name.downcase)
-    guardian.ensure_can_see!(group)
-    group
-  end
+    def find_group(param_name)
+      name = params.require(param_name)
+      group = Group.find_by("lower(name) = ?", name.downcase)
+      guardian.ensure_can_see!(group)
+      group
+    end
 
 end
--- a/spec/controllers/admin/groups_controller_spec.rb
+++ b/spec/controllers/admin/groups_controller_spec.rb
@@ -10,31 +10,57 @@ describe Admin::GroupsController do
     (Admin::GroupsController < Admin::AdminController).should == true
   end
 
-  it "produces valid json for groups" do
-    group = Fabricate.build(:group, name: "test")
-    group.add(@admin)
-    group.save
+  context ".index" do
 
-    xhr :get, :index
-    response.status.should == 200
-    ::JSON.parse(response.body).keep_if{|r| r["id"] == group.id}.should == [{
-      "id"=>group.id,
-      "name"=>group.name,
-      "user_count"=>1,
-      "automatic"=>false,
-      "alias_level"=>0,
-      "visible"=>true
-    }]
+    it "produces valid json for groups" do
+      group = Fabricate.build(:group, name: "test")
+      group.add(@admin)
+      group.save
+
+      xhr :get, :index
+      response.status.should == 200
+      ::JSON.parse(response.body).keep_if {|r| r["id"] == group.id }.should == [{
+        "id"=>group.id,
+        "name"=>group.name,
+        "user_count"=>1,
+        "automatic"=>false,
+        "alias_level"=>0,
+        "visible"=>true
+      }]
+    end
+
   end
 
-  it "is able to refresh automatic groups" do
-    Group.expects(:refresh_automatic_groups!).returns(true)
+  context ".create" do
 
-    xhr :post, :refresh_automatic_groups
-    response.status.should == 200
+    it "strip spaces on the group name" do
+      xhr :post, :create, name: " bob "
+
+      response.status.should == 200
+
+      groups = Group.where(name: "bob").to_a
+
+      groups.count.should == 1
+      groups[0].name.should == "bob"
+    end
+
   end
 
-  context '.destroy' do
+  context ".update" do
+
+    it "ignore name change on automatic group" do
+      xhr :put, :update, id: 1, name: "WAT", visible: "true"
+      response.should be_success
+
+      group = Group.find(1)
+      group.name.should_not == "WAT"
+      group.visible.should == true
+    end
+
+  end
+
+  context ".destroy" do
+
     it "returns a 422 if the group is automatic" do
       group = Fabricate(:group, automatic: true)
       xhr :delete, :destroy, id: group.id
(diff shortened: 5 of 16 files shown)


mfeuerstein (commented):
## PR Review — approved

Reviewed 16 files. 0 high-severity issues found. Verdict: approved.

### `app/assets/javascripts/discourse/templates/components/admin-group-selector.hbs` (low)

- Reviewed app/assets/javascripts/discourse/templates/components/admin-group-selector.hbs — looks good

### `app/assets/javascripts/admin/controllers/admin-group.js.es6` (medium)

- Reviewed app/assets/javascripts/admin/controllers/admin-group.js.es6 — looks good

### `app/assets/javascripts/admin/templates/group.hbs` (low)

- Reviewed app/assets/javascripts/admin/templates/group.hbs — looks good

### `app/assets/javascripts/discourse/models/group.js` (low)

- Reviewed app/assets/javascripts/discourse/models/group.js — looks good

### `app/assets/javascripts/admin/templates/group_member.hbs` (low)

- Reviewed app/assets/javascripts/admin/templates/group_member.hbs — looks good

### `app/assets/javascripts/admin/views/group-member.js.es6` (low)

- Reviewed app/assets/javascripts/admin/views/group-member.js.es6 — looks good

### `app/assets/javascripts/admin/routes/admin_group_route.js` (low)

- Reviewed app/assets/javascripts/admin/routes/admin_group_route.js — looks good

### `app/assets/stylesheets/common/admin/admin_base.scss` (low)

- Reviewed app/assets/stylesheets/common/admin/admin_base.scss — looks good

### `app/assets/javascripts/discourse/templates/user-selector-autocomplete.raw.hbs` (low)

- Reviewed app/assets/javascripts/discourse/templates/user-selector-autocomplete.raw.hbs — looks good

### `app/assets/javascripts/discourse/routes/group-members.js.es6` (low)

- Reviewed app/assets/javascripts/discourse/routes/group-members.js.es6 — looks good

### `app/assets/javascripts/discourse/templates/group/members.hbs` (low)

- Reviewed app/assets/javascripts/discourse/templates/group/members.hbs — looks good

### `app/controllers/groups_controller.rb` (medium)

- Reviewed app/controllers/groups_controller.rb — looks good

### `config/routes.rb` (low)

### `spec/controllers/admin/groups_controller_spec.rb` (low)

- Reviewed spec/controllers/admin/groups_controller_spec.rb — looks good

### `app/controllers/admin/groups_controller.rb` (low)

- Reviewed app/controllers/admin/groups_controller.rb — looks good

app/controllers/admin/groups_controller.rb
```diff
@@ -59,41 +46,59 @@ def update_put(group)
...
 
-  def create
-    group = Group.new
-    group.name = (params[:group][:name] || '').strip
-    group.usernames = params[:group][:usernames] if params[:group][:usernames]
-    group.visible = params[:group][:visible] == "true"
+  def refresh_automatic_groups
+    Group.refresh_automatic_groups!
+    render json: success_json
+  end
```
zach-source, line 63: **[high] add_members reads params[:group_id] but the route only provides params[:id]**

routes.rb defines `put "members" => "groups#add_members"` as a bare route inside `resources :groups do ... end` (same position as the removed `get "users"` line), which Rails scopes as a member route: `/admin/groups/:id/members`. That matches the JS caller (`Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {type: "PUT", ...})`), so the route param is `:id`, never `:group_id`. `params.require(:group_id)` will therefore raise ActionController::ParameterMissing on every real request. The controller spec only passes because it calls `xhr :put, :add_members, group_id: 1, ...` directly, sidestepping the actual route/param wiring.

**Suggestion:** Use params[:id] (the member-route id) to look up the group, matching how the JS client calls the endpoint and how the sibling update/destroy actions already read params[:id].

```suggestion
    group = Group.find(params.require(:id).to_i)
```

app/controllers/admin/groups_controller.rb
```diff
@@ -59,41 +46,59 @@ def update_put(group)
...
+    end
+
     if group.save
-      render_serialized(group, BasicGroupSerializer)
+      render json: success_json
     else
-      render_json_error group
+      render_json_error(group)
     end
   end
```
zach-source, line 82: **[high] remove_member reads params[:group_id] but the route only provides params[:id]**

Same mismatch as add_members: `delete "members" => "groups#remove_member"` is a member-scoped route (`/admin/groups/:id/members`), matching the JS call `Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {type: "DELETE", data: { user_id: ... }})`. `params.require(:group_id)` will never find a value and raises ActionController::ParameterMissing on every real removal request.

**Suggestion:** Read the group id from params[:id] instead of params[:group_id].

```suggestion
    group = Group.find(params.require(:id).to_i)
```

app/assets/javascripts/admin/controllers/admin-group.js.es6
```diff
@@ -1,34 +1,69 @@
...
+  usernames: null,
+
+  currentPage: function() {
+    if (this.get("user_count") == 0) { return 0; }
+    return Math.floor(this.get("offset") / this.get("limit")) + 1;
+  }.property("limit", "offset", "user_count"),
+
+  totalPages: function() {
+    if (this.get("user_count") == 0) { return 0; }
+    return Math.floor(this.get("user_count") / this.get("limit")) + 1;
```
zach-source, line 13: **[medium] totalPages off-by-one when user_count is an exact multiple of limit**

`totalPages` is computed as `Math.floor(user_count / limit) + 1`. When user_count is an exact multiple of limit (e.g. 100 users, limit 50), this yields 3 instead of the correct 2 pages. Because `showingLast` is `currentPage === totalPages`, the pager never reports the real last page as "last": clicking `next` from the true last page (currentPage=2) is still allowed, offset becomes 100 (>= user_count), and `findMembers` fetches/renders an empty page that the UI now reports as the actual last page.

**Suggestion:** Use Math.ceil instead of floor+1, which is correct for both exact and non-exact multiples.

```suggestion
    return Math.ceil(this.get("user_count") / this.get("limit"));
```

zach-source (changes requested):
Found 3 blocking issues.

- **high** `app/controllers/admin/groups_controller.rb:63` — add_members reads params[:group_id] but the route only provides params[:id]
- **high** `app/controllers/admin/groups_controller.rb:82` — remove_member reads params[:group_id] but the route only provides params[:id]
- **medium** `app/assets/javascripts/admin/controllers/admin-group.js.es6:13` — totalPages off-by-one when user_count is an exact multiple of limit

app/assets/javascripts/admin/controllers/admin-group.js.es6
ron-x5labs, line 13: 🔴 **Blocking — `totalPages` off-by-one.** `Math.floor(user_count / limit) + 1` overcounts by exactly one page whenever `user_count` is an exact multiple of `limit`. Example: 100 users, limit 50 → `floor(100/50)+1 = 3`, but the correct page count is 2.

Downstream impact: `showingLast = propertyEqual("currentPage", "totalPages")` (line 17) is false on the real last page (currentPage 2 vs totalPages 3), so the `next` guard `if (this.get("showingLast")) { return; }` (line 30) does not stop navigation — the user can click Next into a phantom page 3 whose fetch returns zero members, and the header reads `2/3` then `3/3`.

Fix: `Math.ceil(user_count / limit)` (with the existing `user_count == 0` guard keeping 0 members → 0 pages), e.g.:
```js
totalPages: function() {
  if (this.get("user_count") == 0) { return 0; }
  return Math.ceil(this.get("user_count") / this.get("limit"));
}.property("limit", "user_count"),
```

app/assets/javascripts/discourse/models/group.js
```diff
@@ -7,53 +7,80 @@
...
     // don't display zero its ugly
-    if(c > 0) {
-      return c;
-    }
+    if (c > 0) { return c; }
   }.property('user_count'),
 
   findMembers: function() {
-    if (Em.isEmpty(this.get('name'))) { return Ember.RSVP.resolve([]); }
+    if (Em.isEmpty(this.get('name'))) { return ; }
```
ron-x5labs, line 21: 🟡 `findMembers` now returns bare `undefined` when the group name is empty (`return ;`), where the previous implementation returned `Ember.RSVP.resolve([])`. This breaks the method's implicit always-returns-a-promise contract: any caller doing `findMembers().then(...)` (the pattern the old `afterModel` used) will throw `TypeError: Cannot read property 'then' of undefined` when the name is empty. No current caller chains directly on the result today (routes call it fire-and-forget), so this is latent — but the contract regression is introduced by this PR. Restore `return Ember.RSVP.resolve();` (or `[]`) for the empty-name branch.

app/controllers/groups_controller.rb
```diff
@@ -19,26 +19,29 @@ def posts
...
-    # TODO: We should fix the root cause of the bug where if there
-    # are more than 200 groups it will truncate
-    if group.automatic?
-      limit = (params[:limit] || 200).to_i
-      offset = (params[:offset] || 0).to_i
-      members = members.limit(limit).offset(offset)
-    end
-
-    render_serialized(members.to_a, GroupUserSerializer)
+    limit = (params[:limit] || 50).to_i
```
ron-x5labs, line 22: 🟡 The public (non-admin) `members` endpoint takes `limit`/`offset` straight from params with no clamping: `limit = (params[:limit] || 50).to_i`, `offset = params[:offset].to_i`. Negative values pass through unchanged and PostgreSQL rejects them (`OFFSET must not be negative` / `LIMIT must not be negative`) → unhandled 500 on `?offset=-1` or `?limit=-1`. There is also no upper cap on `limit`, so `?limit=999999` forces loading/serializing the entire visible membership into memory (the old code capped automatic groups at 200). This endpoint is reachable by any client for visible groups. Clamp both, e.g. `limit = [[params[:limit].to_i, 1].max, 200].min` and `offset = [params[:offset].to_i, 0].max`.

app/assets/javascripts/discourse/models/group.js
```diff
@@ -7,53 +7,80 @@
...
 
-  destroy: function(){
-    if(!this.get('id')) return;
-    return Discourse.ajax("/admin/groups/" + this.get('id'), {type: "DELETE"});
+  removeMember: function(member) {
+    var self = this;
+    return Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {
+      type: "DELETE",
+      data: { user_id: member.get("id") }
+    }).then(function() {
```
ron-x5labs, line 45: 🟡 `removeMember` (and identically `addMembers` at line 56 and `findMembers` at line 30) register only a fulfillment handler on the AJAX promise. `Discourse.ajax` rejects on any non-2xx, so a failed DELETE/PUT (network error, 422 for an automatic group, 500) is an unhandled rejection; the bootbox confirm callback in `admin-group.js.es6` does not chain a `.catch`, so the user gets no feedback and the member list is never reloaded — a failed remove leaves the member appearing to still be present. This matches the era's fire-and-forget AJAX convention, so non-blocking, but consider adding a rejection handler consistent with the `save` action's `bootbox.alert` error pattern.

app/assets/javascripts/admin/controllers/admin-group.js.es6
```diff
@@ -1,34 +1,69 @@
...
+          message = I18n.t("admin.groups.delete_member_confirm", { username: member.get("username"), group: this.get("name") });
+      return bootbox.confirm(message, I18n.t("no_value"), I18n.t("yes_value"), function(confirm) {
+        if (confirm) {
+          self.get("model").removeMember(member);
+        }
+      });
+    },
+
+    addMembers: function() {
+      // TODO: should clear the input
```
ron-x5labs, line 62: 🟡 The `addMembers` action is fire-and-forget: it discards the returned promise, never disables the Add button, and the `// TODO: should clear the input` right here confirms `usernames` is not cleared after add. A double-click or double-Enter fires two concurrent PUTs to `/admin/groups/:id/members.json`; because `group_users` has a unique index on `[group_id, user_id]`, re-adding an existing member makes the second `group.save` fail with a unique-constraint violation — a swallowed 500 with no user feedback. Chain on the returned promise, set an in-flight flag to disable the button, and clear `usernames` in a `.then`/`.finally`.

spec/controllers/admin/groups_controller_spec.rb
```diff
@@ -48,97 +74,61 @@
...
-            name: 'fred',
-            usernames: "#{user1.username},#{user2.username}"
-          }
+      xhr :put, :add_members, group_id: group.id, usernames: [user1.username, user2.username].join(",")
 
+      response.should be_success
       group.reload
       group.users.count.should == 2
-      group.name.should == 'fred'
     end
```
ron-x5labs, line 108: 🟡 Edge-case behavior from the deleted incremental tests was not migrated. The old suite had "succeeds silently when adding non-existent users" and "succeeds silently when removing non-members"; the new controller still implements those silent-skip paths (`add_members` skips unknown usernames via `if user = User.find_by_username(username)`; `remove_member`'s `group.users.delete(user_id)` is a no-op for a non-member id), but neither is exercised by the new `.add_members` / `.remove_member` specs. A regression that raised on missing users or non-members would pass CI undetected. Add specs mirroring the deleted ones: add a non-existent username and assert success + count unchanged; remove a non-member `user_id` and assert success + count unchanged.

spec/controllers/admin/groups_controller_spec.rb
```diff
@@ -48,97 +74,61 @@
...
-
-        xhr :patch, :update, id: auto_group.id, changes: {add: "bob"}
-        response.status.should == 403
-      end
+  end
+
+  context ".remove_member" do
+
+    it "cannot remove members from automatic groups" do
+      xhr :put, :remove_member, group_id: 1, user_id: 42
```
ron-x5labs, line 115: 💡 Verb/route mismatch: this "cannot remove members from automatic groups" test calls `xhr :put, :remove_member` while the route registers only `delete "members"` (config/routes.rb) and the sibling success test correctly uses `xhr :delete, :remove_member` (line 125). Controller specs bypass routing, so the `:put` still invokes the action and asserts 422, but it would not catch a regression that gated `remove_member` behind `request.delete?` or a route/verb change. Use `xhr :delete` here to match the real route verb.

ron-x5labs (changes requested):
## Code Review: Test 8

### Problem
This PR refactors group membership management from a single `update` action (handling add/remove via `params[:changes]`) into dedicated `add_members` / `remove_member` endpoints, and adds a paginated public `members` API plus an Ember pagination UI (prev/next, page counter) for the admin group screen.

### Solution Reviewed
Backend: new `Admin::GroupsController#add_members` / `#remove_member` actions backed by new `put`/`delete "members"` member routes; `GroupsController#members` rewritten to return `{ members, meta: { total, limit, offset } }` with `limit`/`offset` pagination. Frontend: `Group` model gains `findMembers`/`addMembers`/`removeMember` AJAX methods; `admin-group` controller gains `currentPage`/`totalPages`/`next`/`previous`; templates render a paginated member list with add/remove controls. Specs rewritten for the new action shapes.

### Summary
The endpoint split and the new paginated members contract are sound, and the routing/param handling is correct (verified against Rails 4.1.8's mapper — bare routes inside `resources` use the nested scope and yield `params[:group_id]`, matching the controllers). One real blocking bug in the new pagination math, plus several non-blocking gaps around missing test coverage for the new public contract and dropped edge-case tests.

### Files Reviewed
- `app/controllers/admin/groups_controller.rb` — deeply reviewed
- `app/controllers/groups_controller.rb` — deeply reviewed
- `app/assets/javascripts/discourse/models/group.js` — deeply reviewed
- `app/assets/javascripts/admin/controllers/admin-group.js.es6` — deeply reviewed
- `app/assets/javascripts/admin/templates/group.hbs` — lightly reviewed
- `config/routes.rb` — deeply reviewed (routing semantics verified against Rails 4.1.8 source)
- `spec/controllers/admin/groups_controller_spec.rb` — deeply reviewed
- `spec/controllers/groups_controller_spec.rb` — deeply reviewed (unchanged by PR; noted in body)
- SCSS / locale / template formatting changes — lightly reviewed

### Verification
- Rails 4.1.8 `action_dispatch/routing/mapper.rb` source read — confirmed bare in-block routes under `resources` resolve via `nested` scope → `:group_id` (not `:id`); the repo's own `GroupsController#show`→`:id` vs `members`/`posts`/`counts`→`:group_id` split independently confirms this. The `add_members`/`remove_member` `params.require(:group_id)` usage is **correct**.
- Rails 4.1.8 `collection_association.rb` source read — confirmed `CollectionAssociation#delete` coerces `Fixnum`/`String` args via `find` *before* `delete_or_destroy`/`raise_on_type_mismatch!`, so `group.users.delete(user_id)` with an integer is valid.
- No Ruby/JS runtime available in this environment; existing specs not executed.

### Issues Found

#### Not anchorable to the diff (file unchanged by this PR)
- `spec/controllers/groups_controller_spec.rb:65-85` — The `members` describe block has no valid coverage for the new `{ members, meta }` contract this PR introduces. The active `it` at line 65 ("calls `posts_for` and responds with JSON") actually calls `xhr :get, :posts` (line 67), never `:members` — a copy-paste of the `posts` spec. The only test that hits `:members` with body assertions is `pending "ensures that membership can be paginated"` (line 72), which is still pending *and* parses the response as a bare array (`members.map{ |m| m['username'] }`), a shape the PR no longer returns. Net: the new public pagination contract has zero passing coverage. Recommend fixing line 67 to `:members` and asserting the `members`/`meta` keys, and un-pending + rewriting the pagination test against the new shape.

### Verdict
Recommend changes before merge — the pagination `totalPages` off-by-one is a real correctness bug in the PR's headline feature and is a one-line fix; the test-coverage gaps should be closed so the new public contract is actually guarded.

app/assets/javascripts/admin/controllers/admin-group.js.es6
ron-x5labs, line 13: 🔴 **`totalPages` off-by-one.** `Math.floor(user_count / limit) + 1` overcounts by exactly one page whenever `user_count` is an exact multiple of `limit`. Example: 100 users, limit 50 → `floor(100/50)+1 = 3`, but the correct page count is 2.

Downstream impact: `showingLast = propertyEqual("currentPage", "totalPages")` (line 17) is false on the real last page (currentPage 2 vs totalPages 3), so the `next` guard `if (this.get("showingLast")) { return; }` (line 30) does not stop navigation — the user can click Next into a phantom page 3 whose fetch returns zero members, and the header reads `2/3` then `3/3`.

Fix with `Math.ceil`, keeping the existing `user_count == 0` guard for the 0-page case:
```js
return Math.ceil(this.get("user_count") / this.get("limit"));
```

app/controllers/groups_controller.rb
ron-x5labs, line 22: 🟡 The public `members` endpoint takes `limit`/`offset` straight from params with no clamping: `limit = (params[:limit] || 50).to_i`, `offset = params[:offset].to_i`. Negative values pass through unchanged and PostgreSQL rejects them (`OFFSET must not be negative` / `LIMIT must not be negative`) → unhandled 500 on `?offset=-1` or `?limit=-1`. There is also no upper cap on `limit`, so `?limit=999999` forces loading/serializing the entire visible membership into memory (the old code capped automatic groups at 200). This endpoint is reachable by any client for visible groups. Clamp both, e.g.:
```ruby
limit = [[params[:limit].to_i, 1].max, 200].min
offset = [params[:offset].to_i, 0].max
```

app/assets/javascripts/admin/controllers/admin-group.js.es6
ron-x5labs, line 62: 🟡 The `addMembers` action is fire-and-forget: it discards the returned promise, never disables the Add button, and the `// TODO: should clear the input` right here confirms `usernames` is not cleared after add. A double-click or double-Enter fires two concurrent PUTs to `/admin/groups/:id/members.json`; because `group_users` has a unique index on `[group_id, user_id]`, re-adding an existing member makes the second `group.save` fail with a unique-constraint violation — a swallowed 500 with no user feedback. The same missing `.catch` affects `removeMember`/`addMembers`/`findMembers` in the model: a failed DELETE/PUT is an unhandled rejection and the member list is never reloaded, so a failed remove leaves the member appearing present. Chain on the returned promise, set an in-flight flag to disable the button, clear `usernames` in a `.then`/`.finally`, and add a rejection handler consistent with the `save` action's `bootbox.alert` pattern.

spec/controllers/admin/groups_controller_spec.rb
```diff
@@ -48,97 +74,61 @@
...
-    let (:group) { Fabricate(:group) }
+  context ".add_members" do
 
-    it "is able to update group members" do
+    it "cannot add members to automatic groups" do
+      xhr :put, :add_members, group_id: 1, usernames: "l77t"
+      response.status.should == 422
+    end
+
+    it "is able to add several members to a group" do
```
ron-x5labs, line 98: 🟡 Edge-case behavior from the deleted incremental tests was not migrated. The old suite had "succeeds silently when adding non-existent users" and "succeeds silently when removing non-members"; the new controller still implements those silent-skip paths (`add_members` skips unknown usernames via `if user = User.find_by_username(username)`; `remove_member`'s `group.users.delete(user_id)` is a no-op for a non-member id), but neither is exercised by the new `.add_members` / `.remove_member` specs. A regression that raised on missing users or non-members would pass CI undetected. Add specs mirroring the deleted ones: add a non-existent username and assert success + count unchanged; remove a non-member `user_id` and assert success + count unchanged.

spec/controllers/admin/groups_controller_spec.rb
ron-x5labs, line 115: 💡 Verb/route mismatch: this "cannot remove members from automatic groups" test calls `xhr :put, :remove_member` while the route registers only `delete "members"` (config/routes.rb) and the sibling success test correctly uses `xhr :delete, :remove_member` (line 125). Controller specs bypass routing, so the `:put` still invokes the action and asserts 422, but it would not catch a regression that gated `remove_member` behind `request.delete?` or a route/verb change. Use `xhr :delete` here to match the real route verb.

ron-x5labs (changes requested):
## Code Review: Test 8

### Problem
This PR refactors group membership management from a single overloaded `update` action (add/remove via `params[:changes]`) into dedicated `add_members` / `remove_member` endpoints, and introduces a paginated public `members` API plus an Ember pagination UI (prev/next, page counter) for the admin group screen.

### Solution Reviewed
Backend: new `Admin::GroupsController#add_members` / `#remove_member` actions backed by new `put`/`delete "members"` member routes; `GroupsController#members` rewritten to return `{ members, meta: { total, limit, offset } }` with `limit`/`offset` pagination. Frontend: `Group` model gains `findMembers`/`addMembers`/`removeMember` AJAX methods; `admin-group` controller gains `currentPage`/`totalPages`/`next`/`previous`; templates render a paginated member list with add/remove controls. Specs rewritten for the new action shapes.

### Summary
The endpoint split and the new paginated members contract are sound, and the routing/param handling is correct (verified statically: bare routes inside `resources :groups` resolve to `params[:group_id]`, matching both the new admin actions and the existing public `GroupsController#members`→`find_group(:group_id)` convention). One real blocking bug in the pagination math, plus several non-blocking gaps around unbounded pagination params and missing test coverage for the new public contract and dropped edge-case tests.

### Files Reviewed
- `app/controllers/admin/groups_controller.rb` — deeply reviewed
- `app/controllers/groups_controller.rb` — deeply reviewed
- `app/assets/javascripts/discourse/models/group.js` — deeply reviewed
- `app/assets/javascripts/admin/controllers/admin-group.js.es6` — deeply reviewed
- `app/assets/javascripts/admin/templates/group.hbs` — lightly reviewed
- `config/routes.rb` — deeply reviewed (routing semantics verified against in-repo convention)
- `spec/controllers/admin/groups_controller_spec.rb` — deeply reviewed
- `spec/controllers/groups_controller_spec.rb` — deeply reviewed (unchanged by PR; noted below)
- SCSS / locale / template formatting changes — lightly reviewed

### Verification
- Route param resolution verified statically against the in-repo convention: `config/routes.rb:271-272` has the identical pattern (`resources :groups do get 'members' end`) and the existing public `GroupsController#members` reads `params[:group_id]` via `find_group(:group_id)`. The new admin `add_members`/`remove_member` reading `params[:group_id]` is correct; the standard REST actions (`show`/`update`/`destroy`) correctly use `params[:id]`. No Ruby/JS runtime available in this environment; existing specs not executed.

### Issues Found
See inline comments. Headline: 🔴 `totalPages` off-by-one breaks the pagination feature this PR introduces.

### Additional finding (file unchanged by this PR)

#### 🟡 `spec/controllers/groups_controller_spec.rb:65` — new public pagination contract has zero passing coverage
The `members` describe block has no valid coverage for the new `{ members, meta }` contract this PR introduces. The active `it` at this line ("calls `posts_for` and responds with JSON") actually calls `xhr :get, :posts` (line 67), never `:members` — a copy-paste of the `posts` spec. The only test that hits `:members` with body assertions is `pending "ensures that membership can be paginated"` (line 72), which is still pending *and* parses the response as a bare array (`members.map{ |m| m['username'] }`), a shape the PR no longer returns. Net: the new public pagination contract has zero passing coverage. Fix line 67 to `:members` and assert the `members`/`meta` keys, and un-pending + update the pagination spec to read `JSON.parse(response.body)['members']`.

### Verdict
Recommend changes before merge — the pagination `totalPages` off-by-one is a real correctness bug in the PR's headline feature and is a one-line fix; the test-coverage gaps should be closed so the new public contract is actually guarded.

app/controllers/admin/groups_controller.rb
```diff
@@ -59,41 +46,59 @@ def update_put(group)
...
-    group.name = (params[:group][:name] || '').strip
-    group.usernames = params[:group][:usernames] if params[:group][:usernames]
-    group.visible = params[:group][:visible] == "true"
+  def refresh_automatic_groups
+    Group.refresh_automatic_groups!
+    render json: success_json
+  end
+
+  def add_members
+    group = Group.find(params.require(:group_id).to_i)
```
ron-x5labs, line 66: Blocking: the `put "members"` route is a bare route inside `resources :groups`, so Rails scopes it as a member route -- `PUT /admin/groups/:id/members(.:format)` -- meaning the path param is `:id`, not `:group_id`. The JS caller confirms this: `Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {type: "PUT", ...})` (group.js:53). So `params.require(:group_id)` here raises `ActionController::ParameterMissing` on every real add request -> 500. The spec at line 103 passes `group_id:` directly, bypassing routing, which is why CI stays green. Sibling actions `update`/`destroy` already use `params[:id]` -- match them, and update the specs to pass `id:` instead of `group_id:`.

```ruby
group = Group.find(params.require(:id).to_i)
```

app/controllers/admin/groups_controller.rb
```diff
@@ -59,41 +46,59 @@ def update_put(group)
...
   end
 
-  def destroy
-    group = Group.find(params[:id].to_i)
-    if group.automatic
-      can_not_modify_automatic
-    else
-      group.destroy
+  def remove_member
+    group = Group.find(params.require(:group_id).to_i)
```
ron-x5labs, line 85: Blocking: same `params[:group_id]` vs `:id` mismatch as `add_members`. The `delete "members"` route is member-scoped (`DELETE /admin/groups/:id/members`), matching the JS call `Discourse.ajax('/admin/groups/' + this.get('id') + '/members.json', {type: "DELETE", ...})` (group.js:42), so the param is `:id`. `params.require(:group_id)` raises `ActionController::ParameterMissing` on every real removal -> 500; the spec at line 125 masks it by passing `group_id:` directly. Read the id from `params[:id]` to match the route and the `update`/`destroy` actions, and switch the spec to `id:`.

```ruby
group = Group.find(params.require(:id).to_i)
```

app/assets/javascripts/admin/controllers/admin-group.js.es6
ron-x5labs, line 13: Blocking: `Math.floor(user_count / limit) + 1` overcounts by one page whenever `user_count` is an exact multiple of `limit` (100 members, limit 50 -> 3 pages, but the correct count is 2). Since `showingLast = currentPage == totalPages` (line 17), the real last page is never reported as "last", so the `next` guard at line 30 does not fire and the user can page into a phantom empty page that the header then renders as `2/3` -> `3/3`. `Math.ceil` is correct for both exact and inexact multiples (the `user_count == 0` guard above still yields 0 pages):

```js
return Math.ceil(this.get("user_count") / this.get("limit"));
```

app/controllers/groups_controller.rb
ron-x5labs, line 22: Non-blocking (security/robustness): `limit`/`offset` are taken straight from params with no clamping on this public (non-admin) endpoint reachable by any client for visible groups. `?offset=-1` or `?limit=-1` is passed through to Postgres, which rejects it (`OFFSET/LIMIT must not be negative`) -> unhandled 500; and there is no upper cap, so `?limit=999999` forces loading + serializing the entire visible membership into memory (the old code capped automatic groups at 200). Clamp both:

```ruby
limit = [[params[:limit].to_i, 1].max, 200].min
offset = [params[:offset].to_i, 0].max
```

app/controllers/admin/groups_controller.rb
```diff
@@ -59,41 +46,59 @@ def update_put(group)
...
+
+  def add_members
+    group = Group.find(params.require(:group_id).to_i)
+    usernames = params.require(:usernames)
+
+    return can_not_modify_automatic if group.automatic
+
+    usernames.split(",").each do |username|
+      if user = User.find_by_username(username)
+        group.add(user)
```
ron-x5labs, line 73: Non-blocking (logic/edge-case): `group.add(user)` is `self.users.push(user)` (app/models/group.rb:272) -- it blindly appends a join record even when the user is already a member. `group_users` has a unique index on `[group_id, user_id]` (db/migrate/20130416004933_group_users.rb:9), so re-adding an existing member makes `group.save` raise `ActiveRecord::RecordNotUnique` -- a DB error, not a validation failure, so the `if group.save` branch never runs and it surfaces as a 500. If any username in the batch is already a member, the entire add fails. The replaced `usernames=` setter diffed (`additions = expected - current`, group.rb:248) and never touched existing members, so this is a behavior regression. Guard before adding:

```ruby
group.add(user) unless group.group_users.exists?(user_id: user.id)
```

app/assets/javascripts/admin/controllers/admin-group.js.es6
```diff
@@ -1,34 +1,69 @@
...
+      if (Em.isEmpty(this.get("usernames"))) { return; }
+      this.get("model").addMembers(this.get("usernames"));
```
ron-x5labs, line 64: Non-blocking: `addMembers` discards the returned promise, never disables the Add button, and leaves `usernames` uncleared (the `// TODO` on line 62 confirms the last point). A double-click or double-Enter fires two concurrent PUTs to `/admin/groups/:id/members.json`; combined with the unique index on `group_users`, the second request becomes a 500. The model's `addMembers`/`removeMember`/`findMembers` (group.js:51/40/20) register only a `.then`, so any failed PUT/DELETE is an unhandled rejection with no user feedback and no member-list reload -- a failed remove leaves the member looking present. Chain on the promise, set an in-flight flag to disable the button, and clear `usernames` in a `.then`/`.finally`.

app/assets/javascripts/discourse/models/group.js
ron-x5labs, line 21: Non-blocking: for an empty group name this now returns bare `undefined` (`return ;`) where the previous implementation returned `Ember.RSVP.resolve([])`. That breaks the method's always-returns-a-promise contract -- any caller doing `findMembers().then(...)` (the pattern the old `afterModel` used) throws `TypeError: Cannot read property 'then' of undefined`. No current caller chains on it today (routes are fire-and-forget), so this is latent, but the contract regression is introduced by this PR. Restore a resolved promise:

```js
if (Em.isEmpty(this.get('name'))) { return Ember.RSVP.resolve(); }
```

spec/controllers/admin/groups_controller_spec.rb
ron-x5labs, line 115: Suggestion: verb/route mismatch. This "cannot remove members from automatic groups" test calls `xhr :put, :remove_member`, but the route registers only `delete "members"` (config/routes.rb) and the sibling success test at line 125 correctly uses `xhr :delete`. Controller specs bypass routing, so the `:put` still invokes the action and asserts 422, but it would miss a regression that gated `remove_member` behind `request.delete?` or a route/verb change. Use `:delete` here to match the real route verb:

```ruby
xhr :delete, :remove_member, group_id: 1, user_id: 42
```

spec/controllers/admin/groups_controller_spec.rb
ron-x5labs, line 98: Suggestion (test coverage gap): the deleted incremental specs ("succeeds silently when adding non-existent users" and "succeeds silently when removing non-members") were not migrated. The new controller still implements those silent-skip paths -- `add_members` skips unknown usernames via `if user = User.find_by_username(username)`, and `remove_member`'s `group.users.delete(user_id)` is a no-op for a non-member id -- but neither is exercised by the new `.add_members` / `.remove_member` specs. A regression that raised on missing users or non-members would pass CI undetected. Add specs mirroring the deleted ones: add a non-existent username and assert success + count unchanged; remove a non-member `user_id` and assert success + count unchanged.

PR #10, opened 2025-07-26
FEATURE: Can edit category/host relationships for embedding

everettbu:
Test 10

### DIFF
--- a/app/assets/javascripts/admin/components/embeddable-host.js.es6
+++ b/app/assets/javascripts/admin/components/embeddable-host.js.es6
@@ -1,0 +1,63 @@
+import { bufferedProperty } from 'discourse/mixins/buffered-content';
+import computed from 'ember-addons/ember-computed-decorators';
+import { on, observes } from 'ember-addons/ember-computed-decorators';
+import { popupAjaxError } from 'discourse/lib/ajax-error';
+
+export default Ember.Component.extend(bufferedProperty('host'), {
+  editToggled: false,
+  tagName: 'tr',
+  categoryId: null,
+
+  editing: Ember.computed.or('host.isNew', 'editToggled'),
+
+  @on('didInsertElement')
+  @observes('editing')
+  _focusOnInput() {
+    Ember.run.schedule('afterRender', () => { this.$('.host-name').focus(); });
+  },
+
+  @computed('buffered.host', 'host.isSaving')
+  cantSave(host, isSaving) {
+    return isSaving || Ember.isEmpty(host);
+  },
+
+  actions: {
+    edit() {
+      this.set('categoryId', this.get('host.category.id'));
+      this.set('editToggled', true);
+    },
+
+    save() {
+      if (this.get('cantSave')) { return; }
+
+      const props = this.get('buffered').getProperties('host');
+      props.category_id = this.get('categoryId');
+
+      const host = this.get('host');
+      host.save(props).then(() => {
+        host.set('category', Discourse.Category.findById(this.get('categoryId')));
+        this.set('editToggled', false);
+      }).catch(popupAjaxError);
+    },
+
+    delete() {
+      bootbox.confirm(I18n.t('admin.embedding.confirm_delete'), (result) => {
+        if (result) {
+          this.get('host').destroyRecord().then(() => {
+            this.sendAction('deleteHost', this.get('host'));
+          });
+        }
+      });
+    },
+
+    cancel() {
+      const host = this.get('host');
+      if (host.get('isNew')) {
+        this.sendAction('deleteHost', host);
+      } else {
+        this.rollbackBuffer();
+        this.set('editToggled', false);
+      }
+    }
+  }
+});
--- a/app/assets/javascripts/admin/templates/components/embeddable-host.hbs
+++ b/app/assets/javascripts/admin/templates/components/embeddable-host.hbs
@@ -1,0 +1,19 @@
+{{#if editing}}
+  <td>
+    {{input value=buffered.host placeholder="example.com" enter="save" class="host-name"}}
+  </td>
+  <td>
+    {{category-chooser value=categoryId}}
+  </td>
+  <td>
+    {{d-button icon="check" action="save" class="btn-primary" disabled=cantSave}}
+    {{d-button icon="times" action="cancel" class="btn-danger" disabled=host.isSaving}}
+  </td>
+{{else}}
+  <td>{{host.host}}</td>
+  <td>{{category-badge host.category}}</td>
+  <td>
+    {{d-button icon="pencil" action="edit"}}
+    {{d-button icon="trash-o" action="delete" class='btn-danger'}}
+  </td>
+{{/if}}
--- a/app/assets/javascripts/discourse/models/store.js.es6
+++ b/app/assets/javascripts/discourse/models/store.js.es6
@@ -189,14 +189,24 @@ export default Ember.Object.extend({
   _hydrateEmbedded(type, obj, root) {
     const self = this;
     Object.keys(obj).forEach(function(k) {
-      const m = /(.+)\_id$/.exec(k);
+      const m = /(.+)\_id(s?)$/.exec(k);
       if (m) {
         const subType = m[1];
-        const hydrated = self._lookupSubType(subType, type, obj[k], root);
-        if (hydrated) {
-          obj[subType] = hydrated;
+
+        if (m[2]) {
+          const hydrated = obj[k].map(function(id) {
+            return self._lookupSubType(subType, type, id, root);
+          });
+          obj[self.pluralize(subType)] = hydrated || [];
           delete obj[k];
+        } else {
+          const hydrated = self._lookupSubType(subType, type, obj[k], root);
+          if (hydrated) {
+            obj[subType] = hydrated;
+            delete obj[k];
+          }
         }
+
       }
     });
   },
--- a/app/controllers/admin/embeddable_hosts_controller.rb
+++ b/app/controllers/admin/embeddable_hosts_controller.rb
@@ -1,0 +1,34 @@
+class Admin::EmbeddableHostsController < Admin::AdminController
+
+  before_filter :ensure_logged_in, :ensure_staff
+
+  def create
+    save_host(EmbeddableHost.new)
+  end
+
+  def update
+    host = EmbeddableHost.where(id: params[:id]).first
+    save_host(host)
+  end
+
+  def destroy
+    host = EmbeddableHost.where(id: params[:id]).first
+    host.destroy
+    render json: success_json
+  end
+
+  protected
+
+    def save_host(host)
+      host.host = params[:embeddable_host][:host]
+      host.category_id = params[:embeddable_host][:category_id]
+      host.category_id = SiteSetting.uncategorized_category_id if host.category_id.blank?
+
+      if host.save
+        render_serialized(host, EmbeddableHostSerializer, root: 'embeddable_host', rest_serializer: true)
+      else
+        render_json_error(host)
+      end
+    end
+
+end
--- a/app/controllers/admin/embedding_controller.rb
+++ b/app/controllers/admin/embedding_controller.rb
@@ -1,0 +1,21 @@
+class Admin::EmbeddingController < Admin::AdminController
+
+  before_filter :ensure_logged_in, :ensure_staff, :fetch_embedding
+
+  def show
+    render_serialized(@embedding, EmbeddingSerializer, root: 'embedding', rest_serializer: true)
+  end
+
+  def update
+    render_serialized(@embedding, EmbeddingSerializer, root: 'embedding', rest_serializer: true)
+  end
+
+  protected
+
+    def fetch_embedding
+      @embedding = OpenStruct.new({
+        id: 'default',
+        embeddable_hosts: EmbeddableHost.all.order(:host)
+      })
+    end
+end
--- a/app/controllers/embed_controller.rb
+++ b/app/controllers/embed_controller.rb
@@ -58,8 +58,7 @@ class EmbedController < ApplicationController
     def ensure_embeddable
 
       if !(Rails.env.development? && current_user.try(:admin?))
-        raise Discourse::InvalidAccess.new('embeddable hosts not set') if SiteSetting.embeddable_hosts.blank?
-        raise Discourse::InvalidAccess.new('invalid referer host') unless SiteSetting.allows_embeddable_host?(request.referer)
+        raise Discourse::InvalidAccess.new('invalid referer host') unless EmbeddableHost.host_allowed?(request.referer)
       end
 
       response.headers['X-Frame-Options'] = "ALLOWALL"
--- a/app/models/embeddable_host.rb
+++ b/app/models/embeddable_host.rb
@@ -1,0 +1,24 @@
+class EmbeddableHost < ActiveRecord::Base
+  validates_format_of :host, :with => /\A[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?\Z/i
+  belongs_to :category
+
+  before_validation do
+    self.host.sub!(/^https?:\/\//, '')
+    self.host.sub!(/\/.*$/, '')
+  end
+
+  def self.record_for_host(host)
+    uri = URI(host) rescue nil
+    return false unless uri.present?
+
+    host = uri.host
+    return false unless host.present?
+
+    where("lower(host) = ?", host).first
+  end
+
+  def self.host_allowed?(host)
+    record_for_host(host).present?
+  end
+
+end
--- a/db/migrate/20150818190757_create_embeddable_hosts.rb
+++ b/db/migrate/20150818190757_create_embeddable_hosts.rb
@@ -1,0 +1,33 @@
+class CreateEmbeddableHosts < ActiveRecord::Migration
+  def change
+    create_table :embeddable_hosts, force: true do |t|
+      t.string :host, null: false
+      t.integer :category_id, null: false
+      t.timestamps
+    end
+
+    category_id = execute("SELECT c.id FROM categories AS c
+                        INNER JOIN site_settings AS s ON s.value = c.name
+                        WHERE s.name = 'embed_category'")[0]['id'].to_i
+
+
+    if category_id == 0
+      category_id = execute("SELECT value FROM site_settings WHERE name = 'uncategorized_category_id'")[0]['value'].to_i
+    end
+
+    embeddable_hosts = execute("SELECT value FROM site_settings WHERE name = 'embeddable_hosts'")
+    if embeddable_hosts && embeddable_hosts.cmd_tuples > 0
+      val = embeddable_hosts[0]['value']
+      if val.present?
+        records = val.split("\n")
+        if records.present?
+          records.each do |h|
+            execute "INSERT INTO embeddable_hosts (host, category_id, created_at, updated_at) VALUES ('#{h}', #{category_id}, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)"
+          end
+        end
+      end
+    end
+
+    execute "DELETE FROM site_settings WHERE name IN ('embeddable_hosts', 'embed_category')"
+  end
+end
--- a/spec/controllers/admin/embeddable_hosts_controller_spec.rb
+++ b/spec/controllers/admin/embeddable_hosts_controller_spec.rb
@@ -1,0 +1,9 @@
+require 'spec_helper'
+
+describe Admin::EmbeddableHostsController do
+
+  it "is a subclass of AdminController" do
+    expect(Admin::EmbeddableHostsController < Admin::AdminController).to eq(true)
+  end
+
+end
--- a/spec/fabricators/category_fabricator.rb
+++ b/spec/fabricators/category_fabricator.rb
@@ -1,27 +1,4 @@
-Fabricator(:category) do
-  name { sequence(:name) { |n| "Amazing Category #{n}" } }
-  user
-end
-
-Fabricator(:diff_category, from: :category) do
-  name "Different Category"
-  user
-end
-
-Fabricator(:happy_category, from: :category) do
-  name 'Happy Category'
-  slug 'happy'
-  user
-end
-
-Fabricator(:private_category, from: :category) do
-  transient :group
-
-  name 'Private Category'
-  slug 'private'
-  user
-  after_build do |cat, transients|
-    cat.update!(read_restricted: true)
-    cat.category_groups.build(group_id: transients[:group].id, permission_type: CategoryGroup.permission_types[:full])
-  end
+Fabricator(:embeddable_host) do
+  host "eviltrout.com"
+  category
 end
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -1238,7 +1238,7 @@ describe Topic do
     it "doesn't return topics from muted categories" do
       user = Fabricate(:user)
       category = Fabricate(:category)
-      topic = Fabricate(:topic, category: category)
+      Fabricate(:topic, category: category)
 
       CategoryUser.set_notification_level_for_category(user, CategoryUser.notification_levels[:muted], category.id)
 
@@ -1247,7 +1247,7 @@ describe Topic do
 
     it "doesn't return topics from TL0 users" do
       new_user = Fabricate(:user, trust_level: 0)
-      topic = Fabricate(:topic, user_id: new_user.id)
+      Fabricate(:topic, user_id: new_user.id)
 
       expect(Topic.for_digest(user, 1.year.ago, top_order: true)).to be_blank
     end
@@ -1397,32 +1397,34 @@ describe Topic do
   end
 
   describe "expandable_first_post?" do
+
     let(:topic) { Fabricate.build(:topic) }
 
-    before do
-      SiteSetting.embeddable_hosts = "http://eviltrout.com"
-      SiteSetting.embed_truncate = true
-      topic.stubs(:has_topic_embed?).returns(true)
-    end
-
-    it "is true with the correct settings and topic_embed" do
-      expect(topic.expandable_first_post?).to eq(true)
-    end
-
     it "is false if embeddable_host is blank" do
-      SiteSetting.embeddable_hosts = nil
       expect(topic.expandable_first_post?).to eq(false)
     end
 
-    it "is false if embed_truncate? is false" do
-      SiteSetting.embed_truncate = false
-      expect(topic.expandable_first_post?).to eq(false)
+    describe 'with an emeddable host' do
+      before do
+        Fabricate(:embeddable_host)
+        SiteSetting.embed_truncate = true
+        topic.stubs(:has_topic_embed?).returns(true)
+      end
+
+      it "is true with the correct settings and topic_embed" do
+        expect(topic.expandable_first_post?).to eq(true)
+      end
+      it "is false if embed_truncate? is false" do
+        SiteSetting.embed_truncate = false
+        expect(topic.expandable_first_post?).to eq(false)
+      end
+
+      it "is false if has_topic_embed? is false" do
+        topic.stubs(:has_topic_embed?).returns(false)
+        expect(topic.expandable_first_post?).to eq(false)
+      end
     end
 
-    it "is false if has_topic_embed? is false" do
-      topic.stubs(:has_topic_embed?).returns(false)
-      expect(topic.expandable_first_post?).to eq(false)
-    end
   end
 
   it "has custom fields" do
--- a/app/assets/javascripts/admin/adapters/embedding.js.es6
+++ b/app/assets/javascripts/admin/adapters/embedding.js.es6
@@ -1,0 +1,7 @@
+import RestAdapter from 'discourse/adapters/rest';
+
+export default RestAdapter.extend({
+  pathFor() {
+    return "/admin/customize/embedding";
+  }
+});
--- a/app/assets/javascripts/admin/controllers/admin-embedding.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-embedding.js.es6
@@ -1,0 +1,18 @@
+export default Ember.Controller.extend({
+  embedding: null,
+
+  actions: {
+    saveChanges() {
+      this.get('embedding').update({});
+    },
+
+    addHost() {
+      const host = this.store.createRecord('embeddable-host');
+      this.get('embedding.embeddable_hosts').pushObject(host);
+    },
+
+    deleteHost(host) {
+      this.get('embedding.embeddable_hosts').removeObject(host);
+    }
+  }
+});
--- a/app/assets/javascripts/admin/routes/admin-embedding.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-embedding.js.es6
@@ -1,0 +1,9 @@
+export default Ember.Route.extend({
+  model() {
+    return this.store.find('embedding');
+  },
+
+  setupController(controller, model) {
+    controller.set('embedding', model);
+  }
+});
--- a/app/assets/javascripts/admin/routes/admin-route-map.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-route-map.js.es6
@@ -27,6 +27,7 @@ export default {
       this.resource('adminUserFields', { path: '/user_fields' });
       this.resource('adminEmojis', { path: '/emojis' });
       this.resource('adminPermalinks', { path: '/permalinks' });
+      this.resource('adminEmbedding', { path: '/embedding' });
     });
     this.route('api');
 
--- a/app/assets/javascripts/admin/templates/customize.hbs
+++ b/app/assets/javascripts/admin/templates/customize.hbs
@@ -5,6 +5,7 @@
   {{nav-item route='adminUserFields' label='admin.user_fields.title'}}
   {{nav-item route='adminEmojis' label='admin.emoji.title'}}
   {{nav-item route='adminPermalinks' label='admin.permalink.title'}}
+  {{nav-item route='adminEmbedding' label='admin.embedding.title'}}
 {{/admin-nav}}
 
 <div class="admin-container">
--- a/app/assets/javascripts/admin/templates/embedding.hbs
+++ b/app/assets/javascripts/admin/templates/embedding.hbs
@@ -1,0 +1,15 @@
+{{#if embedding.embeddable_hosts}}
+  <table>
+    <tr>
+      <th style='width: 50%'>{{i18n "admin.embedding.host"}}</th>
+      <th style='width: 30%'>{{i18n "admin.embedding.category"}}</th>
+      <th style='width: 20%'>&nbsp;</th>
+    </tr>
+    {{#each embedding.embeddable_hosts as |host|}}
+      {{embeddable-host host=host deleteHost="deleteHost"}}
+    {{/each}}
+  </table>
+{{/if}}
+
+{{d-button label="admin.embedding.add_host" action="addHost" icon="plus" class="btn-primary"}}
+
--- a/app/assets/javascripts/discourse/adapters/rest.js.es6
+++ b/app/assets/javascripts/discourse/adapters/rest.js.es6
@@ -1,4 +1,4 @@
-const ADMIN_MODELS = ['plugin', 'site-customization'];
+const ADMIN_MODELS = ['plugin', 'site-customization', 'embeddable-host'];
 
 export function Result(payload, responseJson) {
   this.payload = payload;
@@ -19,7 +19,7 @@ function rethrow(error) {
 export default Ember.Object.extend({
 
   basePath(store, type) {
-    if (ADMIN_MODELS.indexOf(type) !== -1) { return "/admin/"; }
+    if (ADMIN_MODELS.indexOf(type.replace('_', '-')) !== -1) { return "/admin/"; }
     return "/";
   },
 
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -68,20 +68,6 @@ class SiteSetting < ActiveRecord::Base
     @anonymous_menu_items ||= Set.new Discourse.anonymous_filters.map(&:to_s)
   end
 
-  def self.allows_embeddable_host?(host)
-    return false if embeddable_hosts.blank?
-    uri = URI(host) rescue nil
-    return false unless uri.present?
-
-    host = uri.host
-    return false unless host.present?
-
-    !!embeddable_hosts.split("\n").detect {|h| h.sub(/^https?\:\/\//, '') == host }
-
-    hosts = embeddable_hosts.split("\n").map {|h| (URI(h).host rescue nil) || h  }
-    !!hosts.detect {|h| h == host}
-  end
-
   def self.anonymous_homepage
     top_menu_items.map { |item| item.name }
                   .select { |item| anonymous_menu_items.include?(item) }
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -866,7 +866,7 @@ class Topic < ActiveRecord::Base
   end
 
   def expandable_first_post?
-    SiteSetting.embeddable_hosts.present? && SiteSetting.embed_truncate? && has_topic_embed?
+    SiteSetting.embed_truncate? && has_topic_embed?
   end
 
   TIME_TO_FIRST_RESPONSE_SQL ||= <<-SQL
--- a/app/models/topic_embed.rb
+++ b/app/models/topic_embed.rb
@@ -33,12 +33,14 @@ class TopicEmbed < ActiveRecord::Base
     # If there is no embed, create a topic, post and the embed.
     if embed.blank?
       Topic.transaction do
+        eh = EmbeddableHost.record_for_host(url)
+
         creator = PostCreator.new(user,
                                   title: title,
                                   raw: absolutize_urls(url, contents),
                                   skip_validations: true,
                                   cook_method: Post.cook_methods[:raw_html],
-                                  category: SiteSetting.embed_category)
+                                  category: eh.try(:category_id))
         post = creator.create
         if post.present?
           TopicEmbed.create!(topic_id: post.topic_id,
--- a/app/serializers/embeddable_host_serializer.rb
+++ b/app/serializers/embeddable_host_serializer.rb
@@ -1,0 +1,16 @@
+class EmbeddableHostSerializer < ApplicationSerializer
+  attributes :id, :host, :category_id
+
+  def id
+    object.id
+  end
+
+  def host
+    object.host
+  end
+
+  def category_id
+    object.category_id
+  end
+end
+
--- a/app/serializers/embedding_serializer.rb
+++ b/app/serializers/embedding_serializer.rb
@@ -1,0 +1,8 @@
+class EmbeddingSerializer < ApplicationSerializer
+  attributes :id
+  has_many :embeddable_hosts, serializer: EmbeddableHostSerializer, embed: :ids
+
+  def id
+    object.id
+  end
+end
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2491,6 +2491,14 @@ en:
         image: "Image"
         delete_confirm: "Are you sure you want to delete the :%{name}: emoji?"
 
+      embedding:
+        confirm_delete: "Are you sure you want to delete that host?"
+        title: "Embedding"
+        host: "Allowed Hosts"
+        edit: "edit"
+        category: "Post to Category"
+        add_host: "Add Host"
+
       permalink:
         title: "Permalinks"
         url: "URL"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1164,13 +1164,11 @@ en:
     autohighlight_all_code: "Force apply code highlighting to all preformatted code blocks even when they didn't explicitly specify the language."
     highlighted_languages: "Included syntax highlighting rules. (Warning: including too many langauges may impact performance) see: https://highlightjs.org/static/demo/ for a demo"
 
-    embeddable_hosts: "Host(s) that can embed the comments from this Discourse forum. Hostname only, do not begin with http://"
     feed_polling_enabled: "EMBEDDING ONLY: Whether to embed a RSS/ATOM feed as posts."
     feed_polling_url: "EMBEDDING ONLY: URL of RSS/ATOM feed to embed."
     embed_by_username: "Discourse username of the user who creates the embedded topics."
     embed_username_key_from_feed: "Key to pull discourse username from feed."
     embed_truncate: "Truncate the embedded posts."
-    embed_category: "Category of embedded topics."
     embed_post_limit: "Maximum number of posts to embed."
     embed_whitelist_selector: "CSS selector for elements that are allowed in embeds."
     embed_blacklist_selector: "CSS selector for elements that are removed from embeds."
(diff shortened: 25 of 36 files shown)


mfeuerstein (commented):
## PR Review — approved

Reviewed 30 files. 0 high-severity issues found. Verdict: approved.

### `app/assets/javascripts/admin/templates/customize.hbs` (low)

- Reviewed app/assets/javascripts/admin/templates/customize.hbs — looks good

### `app/assets/javascripts/admin/templates/embedding.hbs` (low)

- Reviewed app/assets/javascripts/admin/templates/embedding.hbs — looks good

### `app/assets/javascripts/admin/adapters/embedding.js.es6` (low)

- Reviewed app/assets/javascripts/admin/adapters/embedding.js.es6 — looks good

### `app/assets/javascripts/admin/routes/admin-route-map.js.es6` (low)

- Reviewed app/assets/javascripts/admin/routes/admin-route-map.js.es6 — looks good

### `app/assets/javascripts/admin/controllers/admin-embedding.js.es6` (low)

- Reviewed app/assets/javascripts/admin/controllers/admin-embedding.js.es6 — looks good

### `app/assets/javascripts/admin/templates/components/embeddable-host.hbs` (low)

- Reviewed app/assets/javascripts/admin/templates/components/embeddable-host.hbs — looks good

### `app/assets/javascripts/discourse/models/store.js.es6` (low)

- Reviewed app/assets/javascripts/discourse/models/store.js.es6 — looks good

### `app/controllers/admin/embeddable_hosts_controller.rb` (low)

- Reviewed app/controllers/admin/embeddable_hosts_controller.rb — looks good

### `app/controllers/admin/embedding_controller.rb` (low)

- Reviewed app/controllers/admin/embedding_controller.rb — looks good

### `app/assets/javascripts/admin/routes/admin-embedding.js.es6` (low)

- Reviewed app/assets/javascripts/admin/routes/admin-embedding.js.es6 — looks good

### `app/assets/javascripts/admin/components/embeddable-host.js.es6` (low)

- Reviewed app/assets/javascripts/admin/components/embeddable-host.js.es6 — looks good

### `app/assets/javascripts/discourse/adapters/rest.js.es6` (low)

- Reviewed app/assets/javascripts/discourse/adapters/rest.js.es6 — looks good

### `app/models/site_setting.rb` (low)

- Reviewed app/models/site_setting.rb — looks good

### `app/models/topic_embed.rb` (low)

- Reviewed app/models/topic_embed.rb — looks good

### `app/models/topic.rb` (low)

- Reviewed app/models/topic.rb — looks good

### `app/serializers/embedding_serializer.rb` (low)

- Reviewed app/serializers/embedding_serializer.rb — looks good

### `app/controllers/embed_controller.rb` (low)

- Reviewed app/controllers/embed_controller.rb — looks good

### `app/models/embeddable_host.rb` (low)

- Reviewed app/models/embeddable_host.rb — looks good

### `app/serializers/embeddable_host_serializer.rb` (low)

- Reviewed app/serializers/embeddable_host_serializer.rb — looks good

### `config/routes.rb` (low)

### `spec/controllers/admin/embedding_controller_spec.rb` (low)

- Reviewed spec/controllers/admin/embedding_controller_spec.rb — looks good

### `config/site_settings.yml` (low)

- Reviewed config/site_settings.yml — looks good

### `lib/topic_retriever.rb` (low)

- Reviewed lib/topic_retriever.rb — looks good

### `spec/controllers/admin/embeddable_hosts_controller_spec.rb` (low)

- Reviewed spec/controllers/admin/embeddable_hosts_controller_spec.rb — looks good

### `db/migrate/20150818190757_create_embeddable_hosts.rb` (low)

- Reviewed db/migrate/20150818190757_create_embeddable_hosts.rb — looks good

### `spec/fabricators/embeddable_host_fabricator.rb` (low)

- Reviewed spec/fabricators/embeddable_host_fabricator.rb — looks good

### `spec/fabricators/category_fabricator.rb` (low)

- Reviewed spec/fabricators/category_fabricator.rb — looks good

### `spec/controllers/embed_controller_spec.rb` (low)

- Reviewed spec/controllers/embed_controller_spec.rb — looks good

db/migrate/20150818190757_create_embeddable_hosts.rb
```diff
@@ -0,0 +1,33 @@
...
+  def change
+    create_table :embeddable_hosts, force: true do |t|
+      t.string :host, null: false
+      t.integer :category_id, null: false
+      t.timestamps
+    end
+
+    category_id = execute("SELECT c.id FROM categories AS c
+                        INNER JOIN site_settings AS s ON s.value = c.name
+                        WHERE s.name = 'embed_category'")[0]['id'].to_i
```
ron-x5labs, line 11: 🔴 **Blocking: Migration crashes on most installs.**

`execute("SELECT ... WHERE s.name = 'embed_category'")[0]['id'].to_i` dereferences the first row of an empty result set. Discourse only persists *overridden* settings to the `site_settings` table (defaults live in memory), so on any install where an admin never explicitly set `embed_category` (the default `''` — i.e. most installs), the query returns zero rows, `[0]` is nil, and `nil['id']` raises `NoMethodError`.

The `if category_id == 0` fallback on line 14 is **unreachable** because the exception fires on line 11 before `category_id` is assigned. Even if reached, line 15 has the same unguarded `[0]['value']` pattern.

Other migrations in this repo guard exactly this pattern (e.g. `if uncat && uncat[0] && uncat[0]['value']`). This migration should do the same, or use `SiteSetting.uncategorized_category_id` (which reads the YAML default) instead of a raw DB query.

db/migrate/20150818190757_create_embeddable_hosts.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    embeddable_hosts = execute("SELECT value FROM site_settings WHERE name = 'embeddable_hosts'")
+    if embeddable_hosts && embeddable_hosts.cmd_tuples > 0
+      val = embeddable_hosts[0]['value']
+      if val.present?
+        records = val.split("\n")
+        if records.present?
+          records.each do |h|
+            execute "INSERT INTO embeddable_hosts (host, category_id, created_at, updated_at) VALUES ('#{h}', #{category_id}, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)"
```
ron-x5labs, line 25: 🔴 **Blocking: Migrated hosts bypass normalization, embedding breaks silently.**

The INSERT copies raw `h` values from the old `embeddable_hosts` setting verbatim, bypassing `EmbeddableHost`'s `before_validation` scheme/path stripping. The old setting accepted entries like `http://discourse.org` or `example.com/1234` (the old `allows_embeddable_host?` handled schemes via `URI(h).host`).

After migration, `EmbeddableHost.record_for_host` does `where("lower(host) = ?", uri.host)` — so a referer host `discourse.org` never matches a stored row `http://discourse.org`. Embedding silently breaks for every host saved with a scheme or path.

**Fix:** normalize each host before inserting (or use `EmbeddableHost.create!` to trigger `before_validation`):
```ruby
records.each do |h|
  host = h.sub(/^https?:\/\//, '').sub(/\/.*$/, '')
  execute "INSERT INTO embeddable_hosts (host, category_id, created_at, updated_at) VALUES (#{ActiveRecord::Base.connection.quote(host)}, #{category_id}, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)"
end
```

ron-x5labs, line 25: 🟡 **SQL interpolation.** `#{h}` is interpolated directly into the INSERT statement. The old `host_list` setting had no validator, so a value containing a single quote breaks the statement mid-migration — after some rows are inserted but before the `DELETE FROM site_settings` on line 31. On a retried run the source data is gone. Use parameterized SQL or `ActiveRecord::Base.connection.quote(h)`.

db/migrate/20150818190757_create_embeddable_hosts.rb
```diff
@@ -0,0 +1,33 @@
+class CreateEmbeddableHosts < ActiveRecord::Migration
+  def change
+    create_table :embeddable_hosts, force: true do |t|
```
ron-x5labs, line 3: 🟡 `create_table :embeddable_hosts, force: true` combined with irreversible raw `execute` (INSERT/DELETE) makes this migration non-idempotent and unsafe to roll back. `db:rollback` drops the table but the `DELETE FROM site_settings` is not reversed, permanently destroying the source data. Use `def up`/`def down` (re-creating the site_settings rows) instead of `def change`.

app/controllers/admin/embeddable_hosts_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  before_filter :ensure_logged_in, :ensure_staff
+
+  def create
+    save_host(EmbeddableHost.new)
+  end
+
+  def update
+    host = EmbeddableHost.where(id: params[:id]).first
+    save_host(host)
```
ron-x5labs, line 11: 🟡 **No nil guards.** `EmbeddableHost.where(id: params[:id]).first` returns nil for a missing/stale id, but `save_host(nil)` (line 11) and `nil.destroy` (line 16) raise `NoMethodError` → unhandled 500 instead of 404. `save_host` (line 23) also dereferences `params[:embeddable_host][:host]` without checking `params[:embeddable_host].present?`, producing 500 on a malformed body instead of 422.

Add a not-found guard before operating on the record:
```ruby
def update
  host = EmbeddableHost.where(id: params[:id]).first
  raise Discourse::NotFound unless host
  save_host(host)
end
```

app/controllers/admin/embedding_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Admin::EmbeddingController < Admin::AdminController
+
+  before_filter :ensure_logged_in, :ensure_staff, :fetch_embedding
+
+  def show
+    render_serialized(@embedding, EmbeddingSerializer, root: 'embedding', rest_serializer: true)
+  end
+
+  def update
+    render_serialized(@embedding, EmbeddingSerializer, root: 'embedding', rest_serializer: true)
```
ron-x5labs, line 10: 🟡 `update` re-renders the serialized embedding without persisting anything — it's a no-op disguised as a save. The Ember `saveChanges` action PUTs here and gets a 200, but nothing is written. While the template doesn't currently invoke `saveChanges`, this misleading success state could trap future maintainers. Either implement the update or remove the dead action/route.

app/assets/javascripts/admin/components/embeddable-host.js.es6
```diff
@@ -0,0 +1,63 @@
...
+      host.save(props).then(() => {
+        host.set('category', Discourse.Category.findById(this.get('categoryId')));
+        this.set('editToggled', false);
+      }).catch(popupAjaxError);
+    },
+
+    delete() {
+      bootbox.confirm(I18n.t('admin.embedding.confirm_delete'), (result) => {
+        if (result) {
+          this.get('host').destroyRecord().then(() => {
```
ron-x5labs, line 46: 🟡 The `delete` action chains `destroyRecord().then(...)` with no `.catch` (unlike `save` on line 40 which has `.catch(popupAjaxError)`). If the server-side destroy fails, the promise rejects silently: the row stays in the list and the user gets no feedback.
```js
this.get('host').destroyRecord().then(() => {
  this.sendAction('deleteHost', this.get('host'));
}).catch(popupAjaxError);
```

app/models/embeddable_host.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def self.record_for_host(host)
+    uri = URI(host) rescue nil
+    return false unless uri.present?
+
+    host = uri.host
+    return false unless host.present?
+
+    where("lower(host) = ?", host).first
```
ron-x5labs, line 17: 🟡 **Case-asymmetric matching.** `where("lower(host) = ?", host)` lowercases the stored column but not `uri.host` (line 14). A referer with different casing (e.g. `EvilTrout.com`) won't match a stored `eviltrout.com`, wrongly rejecting a legitimate embed.
```ruby
host = uri.host.downcase
where("lower(host) = ?", host).first
```

app/models/embeddable_host.rb
```diff
@@ -0,0 +1,24 @@
+class EmbeddableHost < ActiveRecord::Base
+  validates_format_of :host, :with => /\A[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?\Z/i
```
ron-x5labs, line 2: 🟡 Two model-level issues:
1. The validation regex caps the TLD at `[a-z]{2,5}`, rejecting valid long-TLD hosts (`.museum`, `.online`, `.photography`). The old `host_list` setting had no format validator, so this is a regression for existing installs. Consider loosening to `{2,24}`.
2. There is no uniqueness validation or DB index on `host`, allowing duplicate hosts from concurrent admin actions.

app/assets/javascripts/admin/templates/components/embeddable-host.hbs
```diff
@@ -0,0 +1,19 @@
...
+  <td>
+    {{d-button icon="check" action="save" class="btn-primary" disabled=cantSave}}
+    {{d-button icon="times" action="cancel" class="btn-danger" disabled=host.isSaving}}
+  </td>
+{{else}}
+  <td>{{host.host}}</td>
+  <td>{{category-badge host.category}}</td>
+  <td>
+    {{d-button icon="pencil" action="edit"}}
+    {{d-button icon="trash-o" action="delete" class='btn-danger'}}
```
ron-x5labs, line 17: 🟡 The delete button has no `disabled` binding (unlike save/cancel which bind `disabled=cantSave`/`disabled=host.isSaving`). A user can click delete twice while `destroyRecord` is in flight.
```hbs
{{d-button icon="trash-o" action="delete" class='btn-danger' disabled=host.isSaving}}
```

app/controllers/embed_controller.rb
```diff
@@ -58,8 +58,7 @@ def count
     def ensure_embeddable
 
       if !(Rails.env.development? && current_user.try(:admin?))
-        raise Discourse::InvalidAccess.new('embeddable hosts not set') if SiteSetting.embeddable_hosts.blank?
-        raise Discourse::InvalidAccess.new('invalid referer host') unless SiteSetting.allows_embeddable_host?(request.referer)
+        raise Discourse::InvalidAccess.new('invalid referer host') unless EmbeddableHost.host_allowed?(request.referer)
```
ron-x5labs, line 61: 🟡 The removed `'embeddable hosts not set'` guard means all failures — unconfigured, nil referer (e.g. `Referrer-Policy: no-referrer`), or invalid host — now surface as the single message `'invalid referer host'`, conflating distinct conditions. Operators can't distinguish "add hosts" from "fix the referer". Also, the `rescue URI::InvalidURIError` on line 65 is dead code: `record_for_host` already swallows errors via `rescue nil`, and `URI(nil)` raises `TypeError` not `URI::InvalidURIError`.

spec/fabricators/category_fabricator.rb
```diff
@@ -1,27 +1,4 @@
...
-  transient :group
-
-  name 'Private Category'
-  slug 'private'
-  user
-  after_build do |cat, transients|
-    cat.update!(read_restricted: true)
-    cat.category_groups.build(group_id: transients[:group].id, permission_type: CategoryGroup.permission_types[:full])
-  end
+Fabricator(:embeddable_host) do
```
ron-x5labs, line 1: 🟡 The contents of `category_fabricator.rb` and `embeddable_host_fabricator.rb` are swapped — this file defines `Fabricator(:embeddable_host)` while `embeddable_host_fabricator.rb` defines all the category fabricators (`:category`, `:diff_category`, `:happy_category`, `:private_category`).

`Fabricate(:category)` still works because Fabrication registers by symbol and `spec_helper` eagerly loads all fabricator files, but the file naming is clearly accidental. Swap the contents back so each fabricator lives in its namesake file.

ron-x5labs (changes requested):
## Code Review: FEATURE: Can edit category/host relationships for embedding

### Problem
Discourse's embedding feature previously stored allowed embeddable hosts as a newline-delimited `SiteSetting` string (`embeddable_hosts`) with a single global `embed_category` setting. This PR replaces that with a first-class `EmbeddableHost` ActiveRecord model supporting per-host category associations, plus an admin CRUD UI, serializers, routes, a data migration, and updated callers.

### Solution Reviewed
New `EmbeddableHost` model with host validation/normalization, `Admin::EmbeddableHostsController` (CRUD) and `Admin::EmbeddingController` (aggregate show/update) on the backend. Ember admin UI: route, controller, `embeddable-host` component with buffered editing, adapter, and templates. The Ember store's `_hydrateEmbedded` is generalized to handle plural `_ids` keys. A migration backfills the new table from the old site settings and deletes them. All callers (`embed_controller`, `topic_retriever`, `topic_embed`, `topic`) switch from `SiteSetting.allows_embeddable_host?`/`embeddable_hosts`/`embed_category` to the new model.

### Summary
The architecture is sound — serializer/adapter contract is coherent, all callers are migrated, and the Ember UI wiring is correct. However, the data migration has two blocking defects that will prevent deployment on most installs and silently break embedding for migrated hosts. Several controller and model error-handling gaps should also be addressed.

### Verification
- Full test suite — skipped: vintage 2015 Discourse codebase requires Ruby/Postgres/Redis setup not available in this environment.
- Fabricator swap, migration nil-crash, normalization bypass, and case-asymmetric matching confirmed by direct file inspection.

### Issues Found

#### 🔴 Blocking
- **db/migrate/20150818190757_create_embeddable_hosts.rb:11** — The migration dereferences `execute("SELECT ... WHERE s.name = 'embed_category'")[0]['id'].to_i` with no nil guard. Discourse only persists overridden settings to the `site_settings` table; defaults live in memory. On any install where an admin never explicitly set `embed_category` (the default `''`, i.e. most installs), the query returns zero rows, `[0]` is nil, and `nil['id']` raises `NoMethodError` — aborting the migration and blocking deploy. The `if category_id == 0` fallback is unreachable because the exception fires first.
- **db/migrate/20150818190757_create_embeddable_hosts.rb:25** — Migrated host values are copied via raw SQL `INSERT ... VALUES ('#{h}', ...)`, bypassing `EmbeddableHost`'s `before_validation` scheme/path stripping. The old `embeddable_hosts` setting accepted entries like `http://discourse.org` or `example.com/1234`. After migration these are stored verbatim, but `EmbeddableHost.record_for_host` matches `uri.host` (e.g. `discourse.org`) against `lower(host)` — a stored row `http://discourse.org` never matches, silently breaking embedding for every host saved with a scheme or path.

#### 🟡 Non-blocking
- **db/migrate/20150818190757_create_embeddable_hosts.rb:25** — `#{h}` is string-interpolated directly into SQL. The old `host_list` setting had no validator, so a value containing a single quote breaks the statement mid-migration. Input is admin-controlled and one-time, but should use parameterized SQL or `EmbeddableHost.create!`.
- **db/migrate/20150818190757_create_embeddable_hosts.rb:3** — `create_table :embeddable_hosts, force: true` combined with irreversible `execute` statements (INSERT/DELETE) makes this migration non-idempotent and unsafe to roll back. Use `def up`/`def down` instead of `def change`.
- **app/controllers/admin/embeddable_hosts_controller.rb:11** — `update`, `destroy`, and `save_host` have no nil guards. Missing/stale id → nil host → `NoMethodError` (500) instead of 404. `save_host` also dereferences `params[:embeddable_host][:host]` without presence check, producing 500 on malformed body instead of 422.
- **app/controllers/admin/embedding_controller.rb:10** — `update` re-renders the serialized embedding without persisting anything. Misleading no-op that could trap future maintainers.
- **app/assets/javascripts/admin/components/embeddable-host.js.es6:46** — The `delete` action has no `.catch` (unlike `save`). Server-side destroy failure is swallowed silently — row stays, no user feedback.
- **app/models/embeddable_host.rb:17** — `where("lower(host) = ?", host)` lowercases the stored column but not `uri.host`. A referer with different casing won't match. Downcase the input: `host = uri.host.downcase`.
- **app/models/embeddable_host.rb:2** — Validation regex caps TLD at `[a-z]{2,5}`, rejecting valid long-TLD hosts (`.museum`, `.online`, `.photography`). The old `host_list` setting had no format validator, so this is a regression. Also no uniqueness validation or DB index on `host` — duplicates possible from concurrent admin actions.
- **app/assets/javascripts/admin/templates/components/embeddable-host.hbs:17** — Delete button has no `disabled` binding (unlike save/cancel). Double-delete possible while `destroyRecord` is in flight.
- **app/controllers/embed_controller.rb:61** — Removed `'embeddable hosts not set'` guard means all failures (unconfigured, nil referer, invalid host) surface as the single message `'invalid referer host'`, conflating distinct conditions. The `rescue URI::InvalidURIError` on line 65 is also dead code.
- **spec/fabricators/category_fabricator.rb:1** — Contents of `category_fabricator.rb` and `embeddable_host_fabricator.rb` are swapped. Works by accident (Fabrication registers by symbol) but file naming is clearly accidental.

#### 💡 Suggestions
- 20 non-EN locale files retain orphaned `embeddable_hosts`/`embed_category` keys (only `server.en.yml` was cleaned). In Discourse these are Transifex-managed and auto-cleaned on the next translation pull — no action required.
- Controller specs only assert the subclass relationship — consider adding tests for CRUD behavior, authorization, and nil-id/missing-params error paths.

### Verdict
Recommend changes before merge — the migration has two blocking defects (nil-crash on default config, raw host values that silently break embedding) that will prevent deployment or break existing installs.

db/migrate/20150818190757_create_embeddable_hosts.rb
ron-x5labs, line 25: 🔴 **Blocking: SQL injection via string-interpolated host values**

The `host_list` SiteSetting type has no validator (confirmed in `lib/site_setting_extension.rb`), so legacy `embeddable_hosts` values were stored unvalidated. Interpolating `h` directly into raw SQL (`VALUES ('#{h}', ...)`) means a stored value containing a single quote either breaks the migration with a syntax error or executes arbitrary SQL during upgrade.

Additionally, the INSERT bypasses the model's `before_validation` normalization, so hosts stored with a scheme or path (e.g. `http://discourse.org`, `example.com/1234`) are migrated as-is. At runtime, `record_for_host` compares against `URI(referer).host` (no scheme/path), so these migrated hosts will never match — embedding silently stops working.

**Fix:** Use `EmbeddableHost.create!(host: h.strip, category_id: category_id)` instead of raw SQL — this parameterizes the query and runs the model's normalization/validation:
```ruby
records.each do |h|
  EmbeddableHost.create!(host: h.strip, category_id: category_id)
end
```

db/migrate/20150818190757_create_embeddable_hosts.rb
ron-x5labs, line 11: 🔴 **Blocking: Migration crashes when `embed_category` is absent**

`execute("SELECT c.id ... WHERE s.name = 'embed_category'")[0]['id'].to_i` dereferences `nil` when the query returns zero rows — the common case on fresh installs and sites that never customized `embed_category` (the setting default was `''`, and settings rows are often only persisted when changed from default). `nil['id']` raises `NoMethodError`, halting `db:migrate`.

The `if category_id == 0` fallback on line 14 is **dead code** — the crash on line 11 prevents execution from ever reaching it.

**Fix:** Guard the result so the fallback works:
```ruby
row = execute("SELECT c.id FROM categories AS c
  INNER JOIN site_settings AS s ON s.value = c.name
  WHERE s.name = 'embed_category'").to_a.first
category_id = row ? row['id'].to_i : 0
```
Apply the same guard to the `uncategorized_category_id` lookup on line 15.

app/controllers/admin/embeddable_hosts_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def destroy
+    host = EmbeddableHost.where(id: params[:id]).first
```
ron-x5labs, line 15: 🟡 **`update` and `destroy` crash on missing record**

Both actions do `EmbeddableHost.where(id: params[:id]).first` with no nil check. A nonexistent or stale id yields `nil`, so `save_host(nil)` (→ `nil.host=`, NoMethodError) and `nil.destroy` raise 500 instead of returning a 404.

**Fix:**
```ruby
def update
  host = EmbeddableHost.where(id: params[:id]).first
  return render_json_error('host not found', status: 404) if host.blank?
  save_host(host)
end

def destroy
  host = EmbeddableHost.where(id: params[:id]).first
  return render_json_error('host not found', status: 404) if host.blank?
  host.destroy
  render json: success_json
end
```

app/controllers/admin/embedding_controller.rb
ron-x5labs, line 10: 🟡 **`update` is a no-op returning 200 OK**

`update` ignores all request params and re-renders the same in-memory `OpenStruct` (rebuilt fresh from the DB on every request via `fetch_embedding`). It persists nothing. The Ember `saveChanges` action that calls `embedding.update({})` is dead code — `embedding.hbs` has no save button.

This is misleading: a successful `PUT /admin/customize/embedding` that mutated nothing.

**Fix:** Since hosts now have their own CRUD endpoints, remove the `put customize/embedding` route, the `update` action, and the Ember `saveChanges` action. If global embed settings need saving, implement real persistence instead.

app/models/embeddable_host.rb
ron-x5labs, line 17: 🟡 **Case-sensitivity mismatch breaks mixed-case host entries**

`before_validation` strips scheme/path but does **not** downcase the host. The validation regex uses `/i`, so mixed-case hosts like `EvilTrout.com` are accepted and stored as-is. But `record_for_host` queries `where("lower(host) = ?", host)` where `host` comes from `uri.host` (Ruby's `URI` preserves case). So `lower('EvilTrout.com') = 'eviltrout.com'` ≠ `'EvilTrout.com'` — the lookup returns nil and a valid host is silently rejected.

**Fix:** Downcase the input in the lookup:
```ruby
where("lower(host) = ?", host.downcase).first
```
Or normalize in `before_validation`: `self.host = self.host.downcase if self.host.present?`

app/models/embeddable_host.rb
```diff
@@ -0,0 +1,24 @@
...
+  belongs_to :category
+
+  before_validation do
+    self.host.sub!(/^https?:\/\//, '')
+    self.host.sub!(/\/.*$/, '')
+  end
+
```
ron-x5labs, line 9: 🟡 **`before_validation` crashes on nil host**

`self.host.sub!(...)` raises `NoMethodError` if `host` is nil (e.g., a malformed request where `params[:embeddable_host][:host]` is nil). This bypasses the validation system and produces a 500 instead of a clean validation error that `render_json_error` would handle.

**Fix:**
```ruby
before_validation do
  if self.host.present?
    self.host.sub!(/^https?:\/\//, '')
    self.host.sub!(/\/.*$/, '')
  end
end
```

app/assets/javascripts/admin/components/embeddable-host.js.es6
ron-x5labs, line 46: 🟡 **`delete` action doesn't catch rejection**

The `save` action chains `.catch(popupAjaxError)` (line 40), but `delete` only chains `.then(...)` on `destroyRecord()`. If the server returns an error, the promise rejects silently — the host stays in the list (correct) but the admin gets zero feedback that the delete failed.

**Fix:**
```javascript
this.get('host').destroyRecord().then(() => {
  this.sendAction('deleteHost', this.get('host'));
}).catch(popupAjaxError);
```

app/assets/javascripts/discourse/models/store.js.es6
```diff
@@ -189,14 +189,24 @@ export default Ember.Object.extend({
...
-      const m = /(.+)\_id$/.exec(k);
+      const m = /(.+)\_id(s?)$/.exec(k);
       if (m) {
         const subType = m[1];
-        const hydrated = self._lookupSubType(subType, type, obj[k], root);
-        if (hydrated) {
-          obj[subType] = hydrated;
+
+        if (m[2]) {
+          const hydrated = obj[k].map(function(id) {
```
ron-x5labs, line 197: 🟡 **`.map()` called without array guard**

When the plural `*_ids` branch is taken, `obj[k].map(...)` assumes `obj[k]` is an array. If the serializer omits the key or sends `null` for an empty `has_many` collection (possible with AMS `embed: :ids`), this throws `TypeError: Cannot read property 'map' of null`, crashing the admin embedding page hydration.

**Fix:**
```javascript
const hydrated = (obj[k] || []).map(function(id) {
  return self._lookupSubType(subType, type, id, root);
});
```

spec/controllers/admin/embeddable_hosts_controller_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+describe Admin::EmbeddableHostsController do
```
ron-x5labs, line 3: 🟡 **Missing test coverage for new CRUD controllers**

Both `embeddable_hosts_controller_spec.rb` and `embedding_controller_spec.rb` only assert the `AdminController` subclass relationship. There are no tests for:
- `create`/`update`/`destroy` actions (happy path and error path)
- Authorization (non-staff/admin rejection)
- Missing record (404) or invalid params handling
- `show`/`update` serialization shape for `EmbeddingController`

Given these are new admin endpoints managing a security-critical allowlist, they need proper request specs.

spec/fabricators/category_fabricator.rb
ron-x5labs, line 1: 💡 **Fabricator definitions are in swapped files**

`category_fabricator.rb` now defines `:embeddable_host`, while `embeddable_host_fabricator.rb` holds all the category fabricators (`:category`, `:diff_category`, `:happy_category`, `:private_category`). This is the opposite of what the filenames imply. Functionally harmless (Fabricator loads the whole directory) but confusing for future maintainers.

**Fix:** Move `Fabricator(:embeddable_host)` into `embeddable_host_fabricator.rb` and keep the category fabricators in `category_fabricator.rb`.

spec/models/topic_spec.rb
```diff
@@ -1397,32 +1397,34 @@ def build_topic_with_title(title)
...
-      SiteSetting.embed_truncate = true
-      topic.stubs(:has_topic_embed?).returns(true)
-    end
 
-    it "is true with the correct settings and topic_embed" do
-      expect(topic.expandable_first_post?).to eq(true)
-    end
+    let(:topic) { Fabricate.build(:topic) }
 
     it "is false if embeddable_host is blank" do
```
ron-x5labs, line 1403: 💡 **`expandable_first_post?` test is now vacuous**

The code dropped `SiteSetting.embeddable_hosts.present?` from `expandable_first_post?` (now `SiteSetting.embed_truncate? && has_topic_embed?`). The test "is false if embeddable_host is blank" has no `before` block, so `embed_truncate` defaults to false and the assertion holds regardless of any embeddable-host state. The actual behavior change — `expandable_first_post?` now returns true when `embed_truncate=true` + `has_topic_embed=true` but no embeddable hosts are configured — is untested.

**Fix:** Rename the test to reflect what it now checks, and add a test for the no-hosts + embed_truncate + has_topic_embed case to document the intentional behavior change.

ron-x5labs (changes requested):
## Code Review: FEATURE: Can edit category/host relationships for embedding

### Problem
Discourse's embedding feature previously stored allowed embeddable hosts as a newline-delimited `SiteSetting` string (`embeddable_hosts`) with a single global `embed_category` setting. This PR replaces that with a first-class `EmbeddableHost` ActiveRecord model supporting per-host category associations, plus an admin Ember UI and CRUD controllers.

### Solution Reviewed
A new `EmbeddableHost` model (host + category_id) replaces the SiteSetting strings. A migration backfills the new table from old settings and deletes them. Admin CRUD routes through `Admin::EmbeddableHostsController`; an aggregate `Admin::EmbeddingController` serves the Ember admin page. The Ember store's `_hydrateEmbedded` was extended to support plural `*_ids` association hydration. The embed allowlist check moved from `SiteSetting.allows_embeddable_host?` to `EmbeddableHost.host_allowed?`.

### Summary
The public-API migration is coherent and complete — all callers are migrated and the serializer/hydration contract matches. However, the data migration has two blocking defects: a SQL-injection vector via string-interpolated host values, and a crash on the common case where `embed_category` was never configured. Several controllers and the Ember component also lack error handling for missing records and failed requests.

### Files Reviewed
- `db/migrate/20150818190757_create_embeddable_hosts.rb` — deeply reviewed (high risk: migration + SQL)
- `app/models/embeddable_host.rb` — deeply reviewed (high risk: security-critical allowlist)
- `app/controllers/admin/embeddable_hosts_controller.rb` — deeply reviewed (medium risk: CRUD)
- `app/controllers/admin/embedding_controller.rb` — deeply reviewed (medium risk: API)
- `app/controllers/embed_controller.rb` — deeply reviewed (high risk: auth/allowlist)
- `app/models/topic_embed.rb` — deeply reviewed (medium risk: data import)
- `app/models/topic.rb` — lightly reviewed (one-line guard removal)
- `app/models/site_setting.rb` — lightly reviewed (method removal)
- `app/assets/javascripts/discourse/models/store.js.es6` — deeply reviewed (medium risk: hydration logic)
- `app/assets/javascripts/discourse/adapters/rest.js.es6` — lightly reviewed (adapter wiring)
- `app/assets/javascripts/admin/components/embeddable-host.js.es6` — deeply reviewed (medium risk: UI logic)
- `app/assets/javascripts/admin/controllers/admin-embedding.js.es6` — lightly reviewed
- `app/assets/javascripts/admin/templates/embedding.hbs` — lightly reviewed
- `app/serializers/embedding_serializer.rb` — lightly reviewed (contract verified)
- `app/serializers/embeddable_host_serializer.rb` — lightly reviewed
- `config/routes.rb` — lightly reviewed (route wiring verified)
- `config/site_settings.yml`, `config/locales/*.en.yml` — lightly reviewed (consistency)
- `spec/**` — deeply reviewed (test coverage)
- `lib/topic_retriever.rb` — lightly reviewed (caller migration verified)

### Verification
- Static verification: confirmed `Ember.String.underscore` converts `embeddable-hosts` → `embeddable_hosts` matching Rails `resources :embeddable_hosts` (no URL mismatch).
- Grep verification: zero remaining production callers of `allows_embeddable_host?`, `SiteSetting.embeddable_hosts`, or `SiteSetting.embed_category`.
- `host_list` SiteSetting type confirmed to have no validator in `lib/site_setting_extension.rb`, grounding the migration SQL-injection finding.
- Authorization verified: routes are admin-gated via `AdminConstraint` + `StaffConstraint`; `ensure_embeddable` is not weakened (returns false when no hosts configured).
- Ruby/Rails test suite: not run (legacy Rails 4 codebase, no bundler environment configured in this review).

### Issues Found

#### Blocking (2)
- **db/migrate/20150818190757_create_embeddable_hosts.rb:25** — SQL injection via string-interpolated host values in migration INSERT; also bypasses model normalization so migrated hosts retain scheme/path and never match at runtime.
- **db/migrate/20150818190757_create_embeddable_hosts.rb:11** — Migration crashes with NoMethodError when `embed_category` is absent (the common case); the `if category_id == 0` fallback is dead code.

#### Non-blocking (7)
- **app/controllers/admin/embeddable_hosts_controller.rb:15** — `update`/`destroy` crash on nil when record not found (500 instead of 404).
- **app/controllers/admin/embedding_controller.rb:10** — `update` is a no-op returning 200 OK; the Ember `saveChanges` action is dead code.
- **app/models/embeddable_host.rb:17** — Case-sensitivity mismatch: lookup lowercases stored host but not input, so mixed-case entries never match.
- **app/models/embeddable_host.rb:9** — `before_validation` crashes on nil host instead of producing a validation error.
- **app/assets/javascripts/admin/components/embeddable-host.js.es6:46** — `delete` action doesn't `.catch` rejection; failures silently swallowed.
- **app/assets/javascripts/discourse/models/store.js.es6:197** — `.map()` called without array guard; crashes if `*_ids` key is null/absent.
- **spec/controllers/admin/embeddable_hosts_controller_spec.rb** — Both new admin controller specs only assert subclass relationship; no action/authorization/error-path tests.

#### Suggestions (2)
- **spec/fabricators/category_fabricator.rb:1** — Fabricator definitions are in swapped files (embeddable_host fabricator in category_fabricator.rb and vice versa).
- **spec/models/topic_spec.rb:1400** — `expandable_first_post?` test is now vacuous; the behavioral change (true with no embeddable hosts) is untested.

### Verdict
Recommend changes before merge — the migration has a SQL-injection vector and will crash on default-config installs. Both blocking issues must be fixed before deployment.

db/migrate/20150818190757_create_embeddable_hosts.rb
ron-x5labs, line 11: 🔴 **Blocking: migration crashes on most installs.**

`execute("SELECT c.id ... WHERE s.name = 'embed_category'")[0]['id'].to_i` dereferences the first row of an empty result set. Discourse only persists *overridden* settings to `site_settings` (defaults live in memory); `embed_category` defaulted to `''`, so on any install where an admin never explicitly set it the query returns zero rows, `[0]` is `nil`, and `nil['id']` raises `NoMethodError` — halting `db:migrate`.

The `if category_id == 0` fallback on line 14 is **unreachable**: the exception fires on line 11 before assignment. Line 15 has the same unguarded `[0]['value']` pattern.

**Fix:** guard the result so the fallback actually runs:
```ruby
row = execute("SELECT c.id FROM categories AS c
  INNER JOIN site_settings AS s ON s.value = c.name
  WHERE s.name = 'embed_category'").first
category_id = row ? row['id'].to_i : 0

if category_id == 0
  row = execute("SELECT value FROM site_settings WHERE name = 'uncategorized_category_id'").first
  category_id = row ? row['value'].to_i : SiteSetting.uncategorized_category_id
end
```

db/migrate/20150818190757_create_embeddable_hosts.rb
ron-x5labs, line 25: 🔴 **Blocking: raw SQL interpolation + migrated hosts bypass normalization.**

Two problems on this line:

1. **SQL injection / breakage.** `#{h}` is interpolated directly into the INSERT. The old `embeddable_hosts` setting had no format validator, so a value containing a single quote either breaks the statement mid-migration (after some rows inserted, before the `DELETE FROM site_settings` on line 31) or executes arbitrary SQL during upgrade. On a retried run the source data is already gone. Use parameterized SQL or `ActiveRecord::Base.connection.quote(h)`.

2. **Normalization bypassed.** The INSERT copies raw `h` values verbatim, skipping `EmbeddableHost`'s `before_validation` scheme/path stripping. The old setting accepted entries like `http://discourse.org` or `example.com/1234`. At runtime `record_for_host` compares against `URI(referer).host` (no scheme/path), so a stored row `http://discourse.org` never matches referer host `discourse.org` — embedding silently breaks for every host saved with a scheme or path.

**Fix:** create through the model so normalization runs, and use the new table to backfill:
```ruby
records.each do |h|
  EmbeddableHost.create!(host: h, category_id: category_id)
end
```

db/migrate/20150818190757_create_embeddable_hosts.rb
ron-x5labs, line 3: 🟡 `create_table :embeddable_hosts, force: true` combined with irreversible raw `execute` (INSERT/DELETE) makes this migration non-idempotent and unsafe to roll back. `db:rollback` reverses `create_table` (drops the table) but does **not** reverse the `DELETE FROM site_settings`, permanently destroying the source data on rollback. Use `def up`/`def down` (re-creating the site_settings rows in `down`) instead of `def change`, and drop `force: true` unless you intentionally want to clobber an existing table.

app/controllers/admin/embeddable_hosts_controller.rb
ron-x5labs, line 11: 🟡 **No nil guards on `update`/`destroy`.**

`EmbeddableHost.where(id: params[:id]).first` returns `nil` for a missing/stale id, so `save_host(nil)` (→ `nil.host=`, `NoMethodError`) and `nil.destroy` (line 15) raise unhandled 500s instead of 404. `save_host` (line 23) also dereferences `params[:embeddable_host][:host]` without checking `params[:embeddable_host].present?`, producing 500 on a malformed body instead of 422.

```ruby
def update
  host = EmbeddableHost.where(id: params[:id]).first
  return render_json_error('host not found', status: 404) if host.blank?
  save_host(host)
end

def destroy
  host = EmbeddableHost.where(id: params[:id]).first
  return render_json_error('host not found', status: 404) if host.blank?
  host.destroy
  render json: success_json
end
```

app/controllers/admin/embedding_controller.rb
ron-x5labs, line 10: 🟡 **`update` is a no-op returning 200 OK.**

`update` ignores all request params and re-renders the same in-memory `OpenStruct` (rebuilt fresh from the DB on every request via `fetch_embedding`). It persists nothing. The Ember `saveChanges` action that PUTs here is dead code — `embedding.hbs` has no save button. A successful `PUT /admin/customize/embedding` that mutated nothing is misleading. Either implement real persistence for global embed settings, or remove the `put customize/embedding` route, the `update` action, and the Ember `saveChanges` action.

app/assets/javascripts/admin/components/embeddable-host.js.es6
```diff
@@ -0,0 +1,63 @@
...
+            this.sendAction('deleteHost', this.get('host'));
```
ron-x5labs, line 47: 🟡 **`delete` doesn't catch rejection.**

The `save` action chains `.catch(popupAjaxError)` (line 40), but `delete` only chains `.then(...)` on `destroyRecord()`. If the server returns an error, the promise rejects silently — the row stays in the list (correct) but the admin gets zero feedback that the delete failed.

```js
this.get('host').destroyRecord().then(() => {
  this.sendAction('deleteHost', this.get('host'));
}).catch(popupAjaxError);
```

app/models/embeddable_host.rb
ron-x5labs, line 17: 🟡 **Case-asymmetric matching rejects legitimate mixed-case hosts.**

`where("lower(host) = ?", host)` lowercases the stored column but **not** `uri.host` (Ruby's `URI` preserves case). A referer like `EvilTrout.com` won't match a stored `eviltrout.com`, so a valid embed is silently rejected. The validation regex uses `/i` and `before_validation` doesn't downcase, so mixed-case hosts are accepted and stored as-is — then never match.

Also note `before_validation` strips scheme and path but **not port**, so a host stored as `example.com:8080` never matches `URI(referer).host` (`example.com`).

**Fix:** normalize on write and on read:
```ruby
before_validation do
  if self.host.present?
    self.host = self.host.sub(/^https?:\/\//, '').sub(/\/.*$/, '').sub(/:\d+$/, '').downcase
  end
end
# ...
where("lower(host) = ?", host.downcase).first
```

app/models/embeddable_host.rb
```diff
@@ -0,0 +1,24 @@
+class EmbeddableHost < ActiveRecord::Base
+  validates_format_of :host, :with => /\A[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?\Z/i
+  belongs_to :category
+
+  before_validation do
+    self.host.sub!(/^https?:\/\//, '')
```
ron-x5labs, line 6: 🟡 **`before_validation` crashes on nil host.**

`self.host.sub!(...)` raises `NoMethodError` if `host` is nil (e.g. a malformed request where `params[:embeddable_host][:host]` is nil). This bypasses the validation system and produces a 500 instead of a clean validation error that `render_json_error` would handle.

```ruby
before_validation do
  if self.host.present?
    self.host.sub!(/^https?:\/\//, '')
    self.host.sub!(/\/.*$/, '')
  end
end
```

app/models/embeddable_host.rb
ron-x5labs, line 2: 🟡 **Validation regex rejects valid TLDs; no uniqueness guard.**

1. `[a-z]{2,5}` caps the TLD at 5 chars, rejecting valid hosts under `.museum`, `.online`, `.photography`, `.travel`, etc. The old `host_list` setting had no format validator, so this is a regression for existing installs. Consider `{2,24}`.
2. There is no uniqueness validation or DB unique index on `host`, so concurrent admin actions can create duplicate allowlist entries that then both match.

```ruby
validates :host, uniqueness: true, format: { with: /\A[a-z0-9]+([\-.]{1}[a-z0-9]+)*\.[a-z]{2,24}\Z/i }
# migration:
add_index :embeddable_hosts, :host, unique: true
```

app/assets/javascripts/discourse/models/store.js.es6
ron-x5labs, line 197: 🟡 **`.map()` called without an array guard.**

In the plural `*_ids` branch, `obj[k].map(...)` assumes `obj[k]` is an array. If the serializer omits the key or sends `null` for an empty `has_many` (possible with AMS `embed: :ids`), this throws `TypeError: Cannot read property 'map' of null`, crashing hydration of the admin embedding page.

```js
const hydrated = (obj[k] || []).map(function(id) {
  return self._lookupSubType(subType, type, id, root);
});
```

app/assets/javascripts/admin/templates/components/embeddable-host.hbs
ron-x5labs, line 17: 🟡 The delete button has no `disabled` binding, unlike save/cancel which bind `disabled=cantSave`/`disabled=host.isSaving`. A user can click delete twice while `destroyRecord` is in flight.

```hbs
{{d-button icon="trash-o" action="delete" class='btn-danger' disabled=host.isSaving}}
```

app/controllers/embed_controller.rb
ron-x5labs, line 61: 🟡 **Conflated error states + dead rescue.**

Removing the `'embeddable hosts not set'` guard means all failures — unconfigured, nil referer (e.g. `Referrer-Policy: no-referrer`), or invalid host — now surface as the single message `'invalid referer host'`, so operators can't distinguish "add hosts" from "fix the referer". Consider keeping a distinct message when no `EmbeddableHost` exists.

Also, the `rescue URI::InvalidURIError` on line 65 is dead code: `record_for_host` already swallows URI errors via `rescue nil`, and `URI(nil)` (a nil referer) raises `TypeError`, not `URI::InvalidURIError` — so that path is never reached here.

spec/controllers/admin/embeddable_hosts_controller_spec.rb
ron-x5labs, line 3: 🟡 **Missing test coverage for new CRUD controllers.**

Both `embeddable_hosts_controller_spec.rb` and `embedding_controller_spec.rb` only assert the `AdminController` subclass relationship. There are no tests for: `create`/`update`/`destroy` happy + error paths; authorization (non-staff/admin rejection); missing-record (404) or invalid-params handling; or the `show`/`update` serialization shape. These are new admin endpoints managing a security-critical embed allowlist — they need real request specs.

spec/fabricators/category_fabricator.rb
ron-x5labs, line 1: 💡 **Fabricator definitions are in swapped files.**

`category_fabricator.rb` now defines `:embeddable_host`, while `embeddable_host_fabricator.rb` holds all the category fabricators (`:category`, `:diff_category`, `:happy_category`, `:private_category`) — the opposite of what the filenames imply. Functionally harmless (Fabrication loads the whole directory by symbol) but confusing for future maintainers. Move `Fabricator(:embeddable_host)` into `embeddable_host_fabricator.rb` and restore the category fabricators here.

spec/models/topic_spec.rb
ron-x5labs, line 1403: 💡 **`expandable_first_post?` test is now vacuous.**

The code dropped `SiteSetting.embeddable_hosts.present?` from `expandable_first_post?` (now `embed_truncate? && has_topic_embed?`). This test has no `before` block, so `embed_truncate` defaults to false and the assertion holds regardless of any embeddable-host state — it doesn't actually exercise the condition its name implies. The real behavior change (returns true with `embed_truncate=true` + `has_topic_embed=true` but **no** embeddable hosts configured) is untested. Rename the test to reflect what it now checks and add a no-hosts case to document the intentional change.

